PR #354, opened 2014-03-16
Flexible vertical modal position

wichert:
I have a site where my content is in an iframe. When I use a modal in that iframe it is vertically centered in the iframe, but the end result is that the modal is partially below the fold in the outer page.

We worked around this by applying a negative top margin to the modal, but that is somewhat fragile. To handle this kind of situation it would be useful to be able to give a (vertical) positioning hint, for example "top of the page".

### DIFF
--- a/src/pat/modal.js
+++ b/src/pat/modal.js
@@ -8,6 +8,7 @@ define([
     var parser = new Parser("modal");
 
     parser.add_argument("class");
+    parser.add_argument("top", "center");
 
     var modal = {
         name: "modal",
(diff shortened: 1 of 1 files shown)


pilz:
@cornae @wichert The discussion sounds to me that you basically agree to merge that, even though it is not that new anymore. The structure has changed quite since then, but I could easily reproduce that on the current master and merge. Should I? Or should we close this?

wichert:
Seamless iframes are now officially dead, so this is definitely still useful. I'll let @cornae make the final call.

cornae:
Vertical positioning is done entitled with CSS (since we dropped ie8
support) and I overrule the .js applied (and often badly updated) top value
with !importants in the CSS. So I would actually rather remove the entire
vertical position calculation from the .js.
Op do 25 aug. 2016 om 09:11 schreef Alexander Pilz <notifications@github.com

pilz:
@cornae Ok, I hear you saying:
1. Close this PR (now)
2. Remove vertical position calc from modal.js (maintenance job later)
Right?

cornae:
That's my proposal yes.

Op do 25 aug. 2016 om 09:27 schreef Alexander Pilz <notifications@github.com

pilz:
Closing in favor of #467

PR #395, opened 2015-02-05
add body class after patterns loaded from registry

vangheem:


vangheem:
Can I trouble you for a tag also? That way I can depend on this change in mockup/CMFPlone. Thank you.

jcbrand:
Sure, done: https://github.com/Patternslib/Patterns/releases/tag/2.0.8

cornae:
Nice

PR #397, opened 2015-02-25
Fix an example to make it fit the description

fulv:
As it is, the example would make `#content` from the response replace `#content` on the page, plus the other stuff.

I think the point of the example is to show how to replace multiple blocks of content, so I fixed it to make it do what the description says.

### DIFF
--- a/demo/inject/documentation.md
+++ b/demo/inject/documentation.md
@@ -54,7 +54,7 @@ Perhaps inserting a single block of content is not enough for your needs? Don't 
 
     <section id="section">
       <form action="/my/form" class="pat-inject"
-       data-pat-inject="#content && #notice #notices::after">
+       data-pat-inject="#content #section && #notice #notices::after">
         …
       </form>
     </section>


jcbrand:
@fulv I notice now that you added the "in progress" label, should I have waited before merging?

fulv:
No, I think that label was added automatically by github.
Thanks for merging!

cornae:
No Waffle adds that, not Github. If you move it to the column ‘Ready for Merge’ then Waffle should remove the label ‘in progress’ and add the label ‘ready for merge’.

> Op 27 feb. 2015, om 06:03 heeft Fulvio Casali <notifications@github.com> het volgende geschreven:
> 
> No, I think that label was added automatically by github.
> Thanks for merging!
>

fulv:
Interesting, thanks!  In this case, I did not go through waffle at all, so
maybe waffle moves a pull request into the In Progress column automatically
when it's created.

Issue #389, opened 2014-12-26
Extension for injection/modal


cornae:
A regular form has an action attribute, pointing at the desired feedback page that the user gets to see when the submit button is hit. There are also cases in which a form will have multiple submit buttons, each expecting a different feedback page. For instance in case of a wizard that has a decision tree behind it. For these cases the 'formaction' attribute is used on the submit button itself. The value of the formaction attribute overrides the value of the action attribute on the form, but only for that respective submit button. This works in all modern browsers and in IE 10+ without additional JavaScript. 

Pat-inject and pat-modal currently don't seem to respect the formaction attribute. What I currently need is that the injection script also looks if there's a formaction attribute and respect its value. The initial need is limited to a situation with 'pat-modal' on a form. I can however foresee the need for formations in combination with pat-inject too in the near future. 

![fullsizerender](https://cloud.githubusercontent.com/assets/738601/5557133/8c3df9cc-8cf2-11e4-8bad-159a1d3aaf7d.jpg)

I vaguely remember that formactions have been discussed in the passed between @wichert, @chaoflow and myself, but I can't remember if there was any complication with it.

fulv:
@jcbrand: it was mentioned today that you might already have something implemented for Staralliance.  Could you please let me know the status?  I'd like to work on this.

jcbrand:
@fulv whatever I did, was done in Patterns core. I did work to allow pat-subform to submit a form based on a *formaction* button being clicked.

I think if you grep for *formaction* and also look at the commits from 2 to 3 weeks ago you'll see what I did.

jcbrand:
@fulv any update on this? If not, I'd like to take this one for next week's sprint.

fulv:
I've been looking at pat-inject yesterday, and might have something to show
today.

fulv:
@jcbrand, please take a look at my branch https://github.com/Patternslib/Patterns/tree/inject-with-formaction and let me know what you think.  I will work more on it tomorrow.

This is my first attempt at handling `formaction` attributes.  The main thing that I know is missing is handling anchors in the `formaction`.  E.g.: when `formaction` is:

    formaction="/feedback/banner-notifications#documents-cut"

then the source `selector` for all configs should be replaced by `#documents-cut`.

PR #398, opened 2015-03-04
Inject with formaction - completes #389

fulv:
@jcbrand, @cornae

### DIFF
--- a/src/pat/inject.js
+++ b/src/pat/inject.js
@@ -72,7 +72,8 @@ define([
                         $el.on("click.pat-inject", _.onClick);
                     } else if ($el.is("form")) {
                         $el.on("submit.pat-inject", _.onSubmit)
-                        .on("click.pat-inject", "[type=submit]", _.onButtonClick);
+                        .on("click.pat-inject", "[type=submit]", ajax.onClickSubmit)
+                        .on("click.pat-inject", "[type=submit][formaction], [type=image][formaction]", _.onFormActionSubmit);
                     } else if ($el.is(".pat-subform")) {
                         log.debug("Initializing subform with injection");
                     }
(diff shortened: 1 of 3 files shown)


fulv:
connects #389

src/pat/inject.js
```diff
@@ -112,6 +112,24 @@ define([
             _.execute(cfgs, $el);
         },
 
+        onButtonClick: function inject_onButtonClick(ev) {
+            ajax.onClickSubmit(ev); // make sure the submitting button is sent with the form
```
jcbrand: This ``onClickSubmit`` method in the ajax.js pattern is actually a misnomer. It makes it sound as if it submits the form, but instead it merely sets data on the form element.

We should consider renaming it to something like ``setFormData``.

src/pat/inject.js
```diff
@@ -72,7 +72,7 @@ define([
                         $el.on("click.pat-inject", _.onClick);
                     } else if ($el.is("form")) {
                         $el.on("submit.pat-inject", _.onSubmit)
-                        .on("click.pat-inject", "[type=submit]", ajax.onClickSubmit);
+                        .on("click.pat-inject", "[type=submit]", _.onButtonClick);
```
jcbrand: You can differentiate between formaction and non-formaction use-cases here by using different jQuery selectors.

For example:
```
$el.on("submit.pat-inject", _.onSubmit)
    .on("click.pat-inject", "[type=submit]", ajax.onClickSubmit)
    .on("click.pat-inject", "[type=submit][formaction]", _.onButtonClick);
```

The ``onButtonClick`` method can then be simpler and smaller because it doesn't need to do the ``if (formaction)`` check. I would then also rename ``onButtonClick`` to something like ``onFormActionSubmit``.

The ``formaction`` attribute can also be used on buttons of ``type="image"`  (see [here](http://html5doctor.com/html5-forms-introduction-and-new-attributes/#formaction)). So the selector needs to be extended.

So, sticking with the previous example, it would look like this:

```
$el.on("submit.pat-inject", _.onSubmit)
    .on("click.pat-inject", "[type=submit]", ajax.onClickSubmit)
    .on("click.pat-inject", "[type=submit][formaction], [type=image][formaction]", _.onButtonClick);
```

src/pat/inject.js
fulv: I guess it's a matter of interpretation.  When I see `onClickSubmit` I think _when Submit is Clicked_, and **not** _if something is Clicked then Submit_.
Either way, a change like that would require simultaneously changing all calls to `onClickSubmit` in all existing patterns.  Out of the scope of this PR.

src/pat/inject.js
fulv: I like these suggestions, I will make the changes.

fulv:
Ok, I've made the suggested changes, and added a test for `<input type="image">`.

jcbrand:
@fulv Great job, thanks!

fulv:
@jcbrand ok to run `make jsrelease`?

jcbrand:
@fulv: sure

PR #405, opened 2015-06-02
Fixes #403 and #404.

### LINKED ISSUE #403: pat-bumper needs a "side" option
See https://github.com/ploneintranet/ploneintranet.prototype/issues/24

We want to be able to specify which side the element should bump against.  Default to be `top`.
Any combination of  `top right bottom left all` should be allowed.  `all` is equivalent to `top right bottom left`.

### LINKED ISSUE #404: pat-bumper needs to bump on any side even in a scrolling container
With my fix to #401 I addressed this case, but only when the scroll container was the `window` element.

This needs to work also in a "contained" scrolling container.

fulv:
@jcbrand I merged the `_updateStatus` and `_updateContainedStatus` functions into one, since the logic should be the same, save for the type of container.

I fixed the tests accordingly, but currently we are only testing that we bump against the top side, none of the other sides.  How do you suggest to structure additional tests?  Technically, it should be a 3D matrix:
(scrolling container/window) x (top/right/bottom/left) x (all/)
That could get unwieldy.

@cornae this fixes #403 and fixes #404.

### DIFF
--- a/src/pat/bumper/bumper.js
+++ b/src/pat/bumper/bumper.js
@@ -22,6 +22,7 @@ define([
     parser.addArgument("bump-remove");
     parser.addArgument("unbump-add");
     parser.addArgument("unbump-remove", "bumped");
+    parser.addArgument("side", "top");
 
     // XXX Handle resize
     var bumper = {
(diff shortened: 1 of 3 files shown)


jcbrand:
Looks good thanks. Concerning tests, should be relatively quick to just copy and paste the `top` tests and then also check for sides, bottom and all right?

PR #410, opened 2015-09-08
pat-tooltip: add content_html option for source parameter

garbas:
- when content_html is set as source you can pass any html string and that is going to be used inserted in tooltip.

- added ``data-option-value`` option for pat-selectbox pattern

- fixed IE9-11 bug in pat-clone that was happening with textareas and placeholder

/cc @pilz @cornae

[Review on Reviewable](https://reviewable.io/reviews/patternslib/patterns/410)

### DIFF
--- a/src/pat/tooltip/tooltip.js
+++ b/src/pat/tooltip/tooltip.js
@@ -26,7 +26,7 @@ define([
     parser.addArgument("height", "auto", ["auto", "max"]);
     parser.addArgument("trigger", "click", ["click", "hover"]);
     parser.addArgument("closing", "auto", ["auto", "sticky", "close-button"]);
-    parser.addArgument("source", "title", ["auto", "ajax", "content", "title"]);
+    parser.addArgument("source", "title", ["auto", "ajax", "content", "content-html", "title"]);
     parser.addArgument("ajax-data-type", "html", ["html", "markdown"]);
     parser.addArgument("delay", 0);
     parser.addArgument("class");
@@ -35,6 +35,7 @@ define([
     var tooltip = {
         name: "tooltip",
         trigger: ".pat-tooltip",
+        jquery_plugin: true,
 
         count: 0,
 
@@ -262,6 +263,9 @@ define([
                 break;
             case "title":
                 $content=$("<p/>").text(options.title);
+                break;
+            case "content-html":
+                $content = $("<div/>").html(options.content);
                 break;
             case "content":
                 href = $trigger.attr("href");
--- a/src/pat/clone/clone.js
+++ b/src/pat/clone/clone.js
@@ -58,8 +58,21 @@ define("pat-clone",[
             $clone.appendTo(this.$el);
             $clone.children().addBack().contents().addBack().filter(this.incrementValues.bind(this));
             $clone.find(this.options.removeElement).on("click", this.remove.bind(this, $clone));
+
+            // IE BUG : Placeholder text becomes actual value after deep clone on textarea
+            // https://connect.microsoft.com/IE/feedback/details/781612/placeholder-text-becomes-actual-value-after-deep-clone-on-textarea
+            if ($.browser.msie !== undefined) {
+              $(':input[placeholder]', $clone).each(function(i, item) {
+                  var $item = $(item);
+                  if ($item.attr('placeholder') === $item.val()) {
+                    $item.val('');
+                  }
+              });
+            }
+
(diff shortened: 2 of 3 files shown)


cornae:
Rok, Does this mean that pat-tooltip is extended with a new property or some magic on a certain property value? Not sure how to interpret this.

cornae:
I see now in the commit in the updated documentation that an extra value is allowed for the property 'source'. I have a comment and a  question. Comment: Please do not use underscores in values or properties for Patterns. We try to stay close to CSS syntax, in which underscores are not used. Also the word content usually has a different meaning. Best to check with me first before inventing new properties or property values. 

Question: Where do we normally leave the HTML content if the source property is set to `content_html`?

garbas:
@cornae basically is the same as ``content`` optionn but it allows html inside text that you pass ... this is needed for creating a link for tooltip in pat-raptor

cornae:
I see. But why would you need a special property for HTML formatted content. Can't `content` on itself already do that?

garbas:
currently it doesn't and i want sure we want to change this since it might
affect something else. if you want i can also extend current `content` options
to also support html. as you prefer.


Quoting Cornelis G. A. Kolbach (2015-09-08 22:57:04)
> I see. But why would you need a special property for HTML formatted content.
> Can't `content` on itself already do that?
> 
> > Op 8 sep. 2015, om 18:31 heeft Rok Garbas <notifications@github.com> het
> volgende geschreven:
> >
> > @cornae basically is the same as content optionn but it allows html inside
> text that you pass ... this is needed for creating a link for tooltip in
> pat-raptor
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
> 
> 
> —
> Reply to this email directly or view it on GitHub.*
> 


--
Rok Garbas - http://www.garbas.si

cornae:
I think so, what would be possible disadvantages? And what would happen if you put in HTML there now? I had actually always assumed it would just pass it right through.

src/pat/tooltip/tooltip.js
```diff
@@ -26,7 +26,7 @@ define([
     parser.addArgument("height", "auto", ["auto", "max"]);
     parser.addArgument("trigger", "click", ["click", "hover"]);
     parser.addArgument("closing", "auto", ["auto", "sticky", "close-button"]);
-    parser.addArgument("source", "title", ["auto", "ajax", "content", "title"]);
+    parser.addArgument("source", "title", ["auto", "ajax", "content_html", "title"]);
```
pilz: @garbas You are changing content to content_html, doesn't that break any existing use of that argument? Should it be ["auto", "ajax", "content", "content-html", "title"] instead?

cornae: Apart from that I still wonder what the advantage is of an extra value to this property.

pilz: I agree with Rok, we don't know enough about how this is used at the moment so we may break things. We can separate the current task at hand from the work necessary to find out how to unify both into two separate tasks without hurting anybody by adding that second option. We might even decide not to document it for now.

garbas: @pilz good catch. i must have changed that like like 10 times before pushing.
@cornae currently if text is passed as we will insert that text into <p> element ... now with content-html we insert this in <div/> element. 

i do agree content option should be also handling html as well as text, but you know where you've been using it and how much can go wrong.

safe option until then would be to add content-html option

garbas:
heads up ... i have also included pat-selectbox change that we were discussion in the email.

cornae:
pat-select right? :)

src/pat/tooltip/tooltip.js
cornae: Yes, Alexander told me this afternoon that with just content the contents get placed in a <p> tag. I may assume that flow elements such as strong and em and also links may still work then, as they are allowed in a p-tag, right? So for now, since we only want this link, the current content method should be sufficient. Or does it escape the HTML code and thus printing them visible to the user?

garbas:
not sure how patternslib has its conventions in code the file is called selectbox.js and is placed in src/pat/selectbox folder ... but i think we are talking about the same pattern :)

cornae:
Some patterns grew in the wild in the past of which we will taste the fruits of misaligned names and property syntax till eternity. Luckily we are more structured today ;).

garbas:
include ie fix for pat-clone ... also create PR here euphorie/osha.oira/#84

PR #413, opened 2015-09-20
tooltip and inject changes

garbas:
i would love to get some comments what i'm doing is in line with patternslib plans.

[Review on Reviewable](https://reviewable.io/reviews/patternslib/patterns/413)

### DIFF
--- a/src/pat/inject/inject.js
+++ b/src/pat/inject/inject.js
@@ -86,6 +86,7 @@ define([
                     _._initAutoloadVisible($el);
                     break;
             }
+
             log.debug("initialised:", $el);
             return $el;
         },
@@ -219,6 +220,18 @@ define([
                     // ideally go there.
                     cfg.$target.removeClass('is-dirty');
                 }
+
+                // pat-inject is used to populate target in some form and when
+                // Cancel button is presed (this triggers reset event on the
+                // form) you would expect to populate with initial placeholder 
+                var $form = cfg.$target.parents('form')
+                if ($form.size() !== 0 && cfg.$target.data('initial-value') === undefined) {
+                    cfg.$target.data('initial-value', cfg.$target.html());
+                    $form.on('reset', function() {
+                        cfg.$target.html(cfg.$target.data('initial-value'))
+                    })
+                }
+
                 return true;
             });
         },
@@ -318,7 +331,6 @@ define([
             if ((cfg.history === "record") && ("pushState" in history)) {
                 history.pushState({'url': cfg.url}, "", cfg.url);
             }
-            $target.trigger("pat-inject-content-loaded-final");
         },
 
         _afterInjection: function ($el, $injected, cfg) {
--- a/src/pat/inject/inject.js
+++ b/src/pat/inject/inject.js
@@ -39,7 +39,7 @@ define([
 
     var _ = {
         name: "inject",
-        trigger: "a.pat-inject, form.pat-inject, .pat-subform.pat-inject",
+        trigger: ".raptor-ui .ui-button.pat-inject, a.pat-inject, form.pat-inject, .pat-subform.pat-inject",
         init: function inject_init($el, opts) {
             var cfgs = _.extractConfig($el, opts);
             // if the injection shall add a history entry and HTML5 pushState
@@ -318,6 +318,7 @@ define([
             if ((cfg.history === "record") && ("pushState" in history)) {
                 history.pushState({'url': cfg.url}, "", cfg.url);
             }
+            $target.trigger("pat-inject-content-loaded-final");
         },
 
         _afterInjection: function ($el, $injected, cfg) {
--- a/src/pat/inject/inject.js
+++ b/src/pat/inject/inject.js
@@ -214,12 +214,6 @@ define([
                         return false;
                     }
                 }
-                if (cfg.confirm == 'class') {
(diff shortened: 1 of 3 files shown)


jcbrand:
> i would love to get some comments what i'm doing is in line with patternslib plans.

@garbas It would help if you explained what the point of this pull request is.

src/pat/inject/inject.js
```diff
@@ -37,7 +37,7 @@ define([
 
     var _ = {
         name: "inject",
-        trigger: "a.pat-inject, form.pat-inject, .pat-subform.pat-inject",
+        trigger: ".ui-button.pat-inject, a.pat-inject, form.pat-inject, .pat-subform.pat-inject",
         init: function inject_init($el, opts) {
```
jcbrand: The use of `.ui-button` here bothers me. AFAIK this is not a Patternslib class and it's not documented anywhere either. There's nothing intuitive about it.

What kind of markup are you trying to support here?

jcbrand:
The tooltip changes look good. The inject change looks unrelated and would ideally be in a different pull request. Also I'm not completely sure what problem you're trying to resolve there.

jcbrand:
@rok whenever you add new parser arguments, please make sure to also document them in the documentation.md file in the pattern's folder. The options are usually at the end of the document under the **Options Reference** heading.

Please also add a test in `src/pat/tooltips/tests.js`, you can run the tests with `make check`.

src/pat/inject/inject.js
garbas: i would like to use pat-inject on a div.ui-button element from raptor-editor toolbar. i would like to adjust raptor-editor's toolbar because there might be other issues with it.

what would be the best way to move on here?

garbas:
@jcbrand will add the docs. i first want to gather the feedback if i'm doing the thing correctly.

src/pat/inject/inject.js
jcbrand: Ok, so this is Raptor specific.

Is the `.pat-raptor` element perhaps a parent of the `div.ui-button` element? Then the selector could be made `.pat-raptor .ui-button.pat-inject`.

That's nicely descriptive. It's clear that this trigger has to do with pat-raptor and buttons clicked.

If `.pat-raptor` is not available, perhaps another class which indicates that this is relevant to Raptor.

garbas:
@jcbrand i've added documentation for tooltip pattern, i've added ``.raptor-ui`` class to trigger of pat-inject and new commit which fixes #760

src/pat/inject/inject.js
```diff
@@ -206,6 +207,14 @@ define([
...
                 }
 
+                var $form = cfg.$target.parents('form')
+                if ($form.size() !== 0 && cfg.$target.data('initial-value') === undefined) {
+                    cfg.$target.data('initial-value', cfg.$target.html());
+                    $form.on('reset', function() {
+                        cfg.$target.html(cfg.$target.data('initial-value'))
+                    })
+                }
+
```
jcbrand: It's not clear to me what's going on here or why this is needed. I'm guessing it's raptor specific? Please add a comment above explaining why this is necessary.

src/pat/inject/inject.js
```diff
@@ -305,6 +314,7 @@ define([
             if ((cfg.history === "record") && ("pushState" in history)) {
                 history.pushState({'url': cfg.url}, "", cfg.url);
             }
+            $target.trigger("pat-inject-content-loaded-final");
         },
```
jcbrand: The name of this event is misleading. The `pat-inject-content-loaded` event is triggered asynchronously after all images have loaded and might therefore be loaded after this event, so `-final` is not correct.

Why can you not use the existing `pat-inject-content-loaded` event?

src/pat/inject/inject.js
garbas: will add the comment. 

just to explain it here. pat-inject is used to populate target in some ``form`` and when ``Cancel`` button is presed (this triggers ``reset`` event on the form) you would expect to populate with initial placeholder 

example for this is attachements at dashboard in ploneintranet

src/pat/inject/inject.js
garbas: what happens with above event is also that it gets triggered for each image that gets loaded right?

could i use imagesLoaded jquery plugin when there are images to trigger only one ``pat-inject-content-loaded`` event? and in case there is no images we only trigger one ``pat-inject-content-loaded`` event as well.

jcbrand: > could i use imagesLoaded jquery plugin when there are images to trigger only one `pat-inject-content-loaded` event? and in case there is no images we only trigger one `pat-inject-content-loaded` event as well.

Yeah, that's roughly what I had in mind as well, but I looked at the code a bit closer now and see another potential issue.

Triggering an event on `$target` won't work in all cases, because `$target` might be replaced entirely by a new element and might therefore not be in the DOM anymore.

In the `_afterInjection` method the `patterns-injected` event is fired. That looks to me like the right event for you to listen to instead. Agreed?

src/pat/inject/inject.js
jcbrand: Thanks for the explanation, makes sense. Please just present it in such a way that it's clear what's happening. I'd even go so far to put it in a separate method, such as `restoreOnFormReset`.

PR #419, opened 2015-10-17
Add modal-active class to an active pat-modal modal

witekdev:


src/pat/modal/modal.js
```diff
@@ -31,7 +31,7 @@ define([
         _init_inject1: function () {
             var opts = {
                 target: "#pat-modal",
-                "class": "pat-modal" + (this.options["class"] ? " " + this.options["class"] : "")
+                "class": "pat-modal modal-active" + (this.options["class"] ? " " + this.options["class"] : "")
             };
```
jcbrand: Please make `modal-active` the default value of the `class` parser argument, then it will be added automatically above and it doesn't have to be hardcoded.

src/pat/modal/modal.js
```diff
@@ -47,6 +47,7 @@ define([
         },
 
         _init_div1: function () {
+            this.$el.addClass("modal-active");
             var $header = $("<div class='header' />"),
```
jcbrand: And then use `this.options["class"]` here.

src/pat/modal/tests.js
```diff
@@ -55,6 +57,7 @@ define(["pat-modal"], function(pattern) {
                 expect($modal.find(".header .close-panel").length).toBeTruthy();
                 expect($modal.find(".panel-content p").length).toBe(2);
                 expect($modal.find(".panel-content h4").length).toBe(1);
+                expect($modal.hasClass("modal-active")).toBeTruthy();
             });
 
             it("Modal with a form that has the pat-modal CSS class", function() {
```
jcbrand: Please add a test for the `class` parser argument to check that by default `modal-active` is set but that it can be changed to other values.

cornae: Is this related to my request to Alexander to have a class modal-active on body when a modal is active? It's not useful on the modal container anyway.

jcbrand: @cornae yes, it's related to your request.

witekdev: @cornae yes, I see now I misread your email request, thought you meant a class on the body of the modal vs. a class on the "body" tag.

jcbrand:
Thanks!

cornae:
As reported to @jcbrand and @pilz, The class `modal-active` which is supposed to appear on `<body>` when a modal is active, appears already when the page loads. This renders the new feature un-useful. Can you fix this so the class `modal-active` only appears when the modal is active and not before?

PR #421, opened 2015-10-17
add custom event option to toggle pattern, to allow to use it as an r…

MrTango:
…eplacement for mockup pattern


jcbrand:
Thanks!

PR #435, opened 2016-01-24
Fix pat ajax caching issues found on IE11

witekdev:


src/pat/ajax/tests.js
```diff
@@ -1,6 +1,13 @@
 define(["pat-ajax"], function(pattern) {
     var $lab;
 
+    // Monkey patch Date.now so that we dont get current time mismatches while testing
+    var originalDatenow = Date.now;
+    Date.now = function() {
+        originalDatenow();
+        return "123"; // dummy data instead of current time
+    };
+
```
jcbrand: Please use the Jasmine `beforeEach` method here, and also in `afterEach`. In `afterEach` you restore the original function, i.e. ` Date.now = originalDatenow;`.

BTW, why do you call `originalDatenow` in your monkey patch?

src/pat/ajax/ajax.js
```diff
@@ -90,7 +90,13 @@ define([
...
                     success: onSuccess
                 };
-
+            // add nocache url parameter to prevent IE11 from caching.         
+            if ( cfg.url.indexOf("?") == -1 ) {
+                args.url = cfg.url + '?nocache=' + Date.now();
+            }
+            else {
+                args.url = cfg.url + '&nocache=' + Date.now();
+            }
```
jcbrand: I'm a little bit worried about badly formatted URLs, which however do sometimes happen. For example a URL which ends in `?`. Could you please handle that one case?

src/pat/ajax/tests.js
witekdev: ok, Il'l fix it. re: originalDatenow - make check was complaining when that I was declaring it without calling it.

cornae:
Is this a pure technical fix or are there also changes to properties that a regular user sets in the HTML?

src/pat/ajax/tests.js
jcbrand: > make check was complaining when that I was declaring it without calling it.

If you restore `Date.now` in `afterEach`, then JSHint won't complain anymore. Just make sure to declare the variable in the scope outside of `beforeEach` and `afterEach`.

jcbrand:
> Is this a pure technical fix or are there also changes to properties that a regular user sets in the HTML?

Currently this is purely a technical measure to bypass caching of HTML fetched via AJAX in IE.

Perhaps we want to make this behavior configurable? I'm not sure yet. It might be preferable to in some cases allow caching. I think for now I would not make it configurable. We still have the option to do that if the need arises.

cornae:
Ok!

Op ma 25 jan. 2016 om 13:53 schreef JC Brand <notifications@github.com>:

jcbrand:
Turns out this isn't necessary after all and that caching can be worked around via `jQuery.ajaxSettings`

PR #443, opened 2016-06-21
Cleanup core/utils.js dependencies

thet:
Remove unused jquery.browser and underscore dependencies from core/utils.js.

I suspect the ``jquery.browser`` dependency being the cause, why patternslib cannot be built on CMFPlone, where we do not include it.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -1,8 +1,8 @@
 define([
     "jquery",
-    "jquery.browser",
-    "underscore"
-], function($) {
+    "underscore",
+    "jquery.browser"  // adds itself to the jquery object, no need to pass to the define callback.
+], function($, _) {
 
     $.fn.safeClone = function () {
         var $clone = this.clone();


jcbrand:
Both dependencies are necessary and are used within that file.

You can search for `_.reduce` (underscore) and `$.browser.msie` (jquery.browser).

thet:
True. Funny thing is, the tests pass. Might be, because it's used elsewhere.
However,
1) I'll update the pull request to explicitly pass ``_`` to the define function for underscore and add a comment for jquery.browser, which adds itself to the jquery object to make that clear.
2) I will add ``jquery.browser`` to CMFPlone's registry.xml.

thet:
@jcbrand looks like underscore methods are copied and included in utils.js. Why that, if we already depend on underscore?
https://github.com/Patternslib/Patterns/blob/b62bb6c6fbf42494576bb5bde1310c1ee9265d5b/src/core/utils.js#L147

thet:
I have updated this PR.

jcbrand:
> looks like underscore methods are copied and included in utils.js. Why that, if we already depend on underscore?

Legacy. The copied method was there first, and then later underscore was introduced. I think I one time tried to remove it, but something broke (because the underscore method evolved further), so I left it.

PR #448, opened 2016-07-25
Allow to configure different data-pat-inject per formaction, so that …

pilz:
…different targets can be configured per formaction.

This solves a longstanding issue where you can configure different formactions, but the response of these formactions must be always the same because you can only have one data-pat-inject - the one on the form itself.

This change allows to override the data-pat-inject on the button, or even on a node in between, to configure custom sources and targets.

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -5,6 +5,7 @@
 - A fix for pat-scroll to scroll up to current scroll container instead of body. 
 - A fix for pat-scroll to await loading of all images before determining the amount to scroll up.
 - A fix for IE10/11 where the modal wouldn`t close anymore due to activeElement being undefined
+- Allow to configure different data-pat-inject per formaction, so that different targets can be configured per formaction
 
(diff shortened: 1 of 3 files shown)


cornae:
I wanted to ask to wait with the merge of this ticket but it already happened. I might not understand it completely but it looks like a functional change and not a bug fix. Functional changes and additions should IMO always be discussed from both the technical and the design perspective, with injection in particular. I cannot remember a discussion or a ticket in this direction and I also don't see any references in that direction, nor do I understand from this merge request itself what is is for. Can we discuss please?

pilz:
Yes, of course. When do you have time? Tonight? I could around 21.

cornae:
Tonight is good, still in France. I should arrive home at 21:00 precisely, so a bit after 21 with a coffee would be great :)

pilz:
perfect

PR #452, opened 2016-08-09
Add repeat polling to pat-inject, as a poor man's substitute to push …

gyst:
In the absence of proper websocket support, this PR makes it possible to craft a repeat poll for updates against the server by using a delayed autoload. Which becomes recursive if the server response again contains a delayed autoload.

It's beyond me to verify the delayed autoload in the test suite, but I've supplied a working manual verification on the pat-inject index page.


pysailor:
Hi JC, could you please take a look if this is ok? Ping me or Guido for background :)

cornae:
Looks like there is a 'ui' change made to a stock pattern, which I always like to be involved in to guard overview and consistency. I see a new value to the trigger property is introduced 'auto-delayed' and a new property 'delay'. I was wondering what the property value 'auto-delayed' adds, if a scenario `delay: ` in combination with any trigger would cause a delay to the injection. Therefor also to `trigger: auto`.

jcbrand:
@pysailor: can you or @gyst provide the background here?

gyst:
The `delay` property only has effect if `trigger: auto-delayed` is chosen. This is done by purpose in order to not suggest that any other trigger can be delayed. I could've added the delay to `trigger: autoload` but I rather did not want to mess with that code path - and if it's there on `autoload` why would it not be there on `autoload-visible` or `click`? Adding delays to all of those is out of my scope.

cornae:
If that is the underlaying argumentation then I would suggest to implement it with the property `delay` which is only supported with autoload. Is that possible? Additional support can always be added later, syntax can NEVER be changed anymore. We had our lessons concerning that already in the past :). 

For the syntax it's also not that strange that certain combinations work and others don't. Just like `height: 100px` works with CSS in combination with `display: block` and not with `display: inline`. It's part of the standard.

gyst:
OK I will do that.

gyst:
Running `make serve` I'm blocked by:

```
bower                        ECONFLICT Unable to find suitable version for get-size
make: *** [stamp-bower] Error 1

```

jcbrand:
@gyst `bower info get-size` shows the available releases for `get-size. You can try downgrading the version or otherwise playing around in the bower.json file.

gyst:
Thanks JC. Turns out a `make clean all` fixed the conflicting version requirements.

gyst:
@cornae I've changed the API now.

cornae:
Is the documentation updated with the new property?

gyst:
@cornae it is now :-)

PR #458, opened 2016-08-23
Documentation for colour-picker

janga1997:
I would like any suggestions. I would also like, if there is, a reason for choosing Spectrum to base colour-picker upon.

### DIFF
--- a/src/pat/colour-picker/colour-picker.css
+++ b/src/pat/colour-picker/colour-picker.css
@@ -1,0 +1,8 @@
+.checkNative , .checkPattern {
+	font-size:3vh;
+}
+
+#native{
+	width: 5%;
+	height:4vh;
+}
--- a/src/pat/colour-picker/colour-picker.js
+++ b/src/pat/colour-picker/colour-picker.js
@@ -6,7 +6,8 @@
  */
 define([
     "pat-registry",
-    "spectrum"
+    "spectrum",
+    "jquery"
 ], function(registry) {
     var _ = {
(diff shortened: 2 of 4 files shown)


pilz:
Sorry, I don't know about the decision for spectrum. Sorry, I don't know about the decision for spectrum.

jcbrand:
Thanks @janga1997, from what I can tell it looks good! I also don't know why spectrum was chosen, I suppose @cornae made that decision.

If we don't find a definitive answer, I propose you remove that part of the documentation again.

cornae:
To be honest I'm not really why specifically Spectrum was chosen at the time. In general, simplicity of markup and not being depended on image files play important roles with these decisions but I'm don't know what was the case here. Would you prefer another one?

janga1997:
@cornae I don't prefer anything. I just wanted to know if there was a reason to choose spectrum, if not, well, I would just leave out that part of the documentation.
Do you think it's a good idea to include the reasons you gave in the documentation?

jcbrand:
@janga1997 sometimes it might be a good idea to explain why a certain library was chosen, but for now I think you can leave them out.

jcbrand:
@janga1997 Can you make another commit to this branch to remove the `### Why did we base this library on Spectrum?` heading?

janga1997:
@jcbrand I opened up a separate pull request. For some reason I do not have the privileges to push a branch directly to Patternslib. Maybe it has something to do with me not being a member?

PR #459, opened 2016-08-24
Fix masonry layouting

pilz:
https://github.com/Patternslib/Patterns/commit/e243605ae097ef45cee065c2ae4a95794aedb937 breaks because the masonry-ready class is now only set if imgLoad actually throws a done. If all images are cached, this doesn't happen. I would like to actually call this.layout simply twice, instead of doing only this.msnry.layout() in line 48


pilz:
invalid

PR #461, opened 2016-08-24
Trigger a relayout when an update event is triggered by pat-switch et al

pilz:
If e.g. a pat-switch changes the amount of space available to or within the masonry, it needs to layout again. Listening to the pat-update event does this.


thet:
+1 + sorry for not giving a reaction sooner.

pilz:
Thanks for the feedback! Glad to hear that you don't have any objections.

PR #465, opened 2016-08-26
Completely redo the scrolling algorithm.

gyst:
**edit: the description below is now superseded. See tail of comments **

> 
> **edit** Actually the backward compatibility injection *does* work so no API change!
> 
> The actual calculation is difficult and requires special markup wich is auto-added and also auto-removed again.
> 
> Immediately *within* the scrollable, around all the scrollable content,
> we need a special wrapper div that provides a positioning context
> by being position:relative. Let's call this .inner-scrollable.
> 
> This enables us to measure the scrolling distance between the #target
> and the top of the scrollable, represented by .inner-scrollable.
> We cannot measure directly against the scrollable itself, because if
> the scrollable is scrolled down, there's invisible (overflowed) content
> 'sticking out' *above* the top of the scrollable.
> 
> Using the .inner-scrollable top enables us to measure this top overflow.
> It doesn't matter whether the scrollable itself provides a positioning
> context, since we cannot use that in our calculations anyway.
> 
> Because we use .position() which measures the offset versus the positioning
> context, we need to walk up the dom across all nested positioning contexts
> until we reach the (logical) top of the scrollable: .inner-scrollable.
> We MUST ignore the offset of .inner-scrollable versus the scrollable itself!!!
>

### DIFF
--- a/src/pat/scroll/debug.html
+++ b/src/pat/scroll/debug.html
@@ -6,7 +6,6 @@
     	<link rel="stylesheet" href="/style/common.css" type="text/css">
         <script src="/main.js" type="text/javascript" charset="utf-8"></script>
 		<script data-main="/src/pat/main" src="/src/bower_components/requirejs/require.js" type="text/javascript" charset="utf-8"></script>
-	</head>
     <style>
         .current {
             background-color: beige;
@@ -22,23 +21,42 @@
         .weird:after {
            content: " [weird!]"
         }
+        .high {
+          height: 100px;
+          background-color: #ddd;
+        }
         article {
             width: 70%;
             height: 25em;
             overflow: auto;
             position: absolute;
-            top: 2em;
-            right: 0;
+            top: 50px;
+           right: 0;
+        border: solid blue 1px;
+           padding: 5px;
         }
         aside {
             width: 29%;
             position: fixed;
-            top: 2em;
+            top: 5em;
             left: 1em;
         }
+        .wrapper {
+          position: relative;
+          border: dotted red 1px;
+          padding: 5px;
+        }
     </style>
+
+	</head>
 	<body>
-          <article class="pat-rich">
+          <article class="pat-rich" id="scrollable">
+            <div class="wrapper"><!-- inner scrollable with position:relative -->
+            <h1>Some more</h1>
+            <p class="high">A lot of stuff above top</p>
+            <h2>.. and more</h2>
+            <div class="wrapper">
+            <p class="high">A lot of stuff above top</p>
             <p id="top">
               <a href="#top" class="pat-scroll">Top</a> |
               <a href="#p1" class="pat-scroll">001</a> |
@@ -126,6 +144,12 @@
               <a href="#bottom" class="pat-scroll">Bottom</a> |
               <a href="#bottom" class="pat-scroll" data-pat-scroll="trigger: auto">Bottom autoscroll</a>
             </p>
+            <p class="high">some extra stuff</p>
+            <p class="high">some extra stuff</p>
+            <p class="high">some extra stuff</p>
+            </div><!-- /wrapper -->
+            </div><!-- /wrapper -->
+            
         </article>
           <aside>
             <h2>Scroll navigation</h2>
--- a/src/pat/scroll/debug.html
+++ b/src/pat/scroll/debug.html
@@ -51,7 +51,6 @@
 	</head>
 	<body>
           <article class="pat-rich" id="scrollable">
-            <div class="wrapper"><!-- inner scrollable with position:relative -->
             <h1>Some more</h1>
             <p class="high">A lot of stuff above top</p>
             <h2>.. and more</h2>
@@ -147,7 +146,6 @@
             <p class="high">some extra stuff</p>
             <p class="high">some extra stuff</p>
             <p class="high">some extra stuff</p>
-            </div><!-- /wrapper -->
             </div><!-- /wrapper -->
             
         </article>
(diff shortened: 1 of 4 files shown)


gyst:
@pilz This happens to already work flawlessly on the *Berkeley New Music Project*.

cornae:
This sounds rather complicated to design with. Markup requirements must for
behaviour is something we try to avoid in Patterns. Also doesn't feel to me
like it's necessary, but of course I might overlook something, but it;s
also possible that there is a bias somewhere. Is it an idea to figure out a
solution/strategy before anything is programmed? I'm not happy to accept
this merge request as it imposes a markup requirement — in particular a
wrapper element, which I find nasty —  and it's not properly backwards
compatible.

Op vr 26 aug. 2016 om 21:32 schreef Guido Stevens <notifications@github.com

gyst:
FYI this is the backward compatibility injection I tried which did not work out.

```
--- a/src/pat/scroll/scroll.js
+++ b/src/pat/scroll/scroll.js
@@ -142,10 +142,7 @@ define([
                 // Immediately *within* the scrollable, around all the scrollable content,
                 // we need a special wrapper div that provides a positioning context
                 // by being position:relative. Let's call this .inner-scrollable.
-                $(scrollable).wrapInner(function() {
-                    return "<div class='inner-scrollable' stype='position:relative'></div>";
-                })
-                var inner_scrollable = $(scrollable).children('.inner-scrollable')[0]
+                var inner_scrollable = $(scrollable).children()[0];
                 // This enables us to measure the scrolling distance between the #target
                 // and the top of the scrollable, represented by .inner-scrollable.
                 // We cannot measure directly against the scrollable itself, because if

```

gyst:
@cornae This is a programming problem so needs to be solved by exploring algorithms that work in many situations.

The original pat-scroll only worked when the scrollable was the document itself. That broke the messages app, which has an overflowing scrollable which is not the document itself. I fixed that and this has become master.

Now Alex found that the new behavior also breaks when there's nested positioning contexts. This PR fixes that. However it then also requires that there's a valid toplevel positioning context that coincides with the (logical, not visual) top of the scrollable. Else we cannot correctly derive the position of the target where we want to scroll to.

gyst:
Actually the backward compatibility injection does work so no API change! I made a silly typo.
I hope you can live with an auto-added div @cornae that provides the required positioning context within the scrollable and does nothing else.

cornae:
I cannot live with an automatically added DIV. It should be a last resort for several reasons, one of which is that designs could break. I can live with it even less this solution is being built without thinking it through before together. Historically we have reached good results on Patterns that require complex scripting and operate in complex CSS layouts when working together with a simple designer.

gyst:
@cornae I've documented all my findings above and made the challenges reproducible in the http://localhost:4001/src/pat/scroll/debug.html debug page on this branch. I've been working on this all day, plus half a day last week. I welcome any suggestions to solve this in a simpler way. Well, once I'm fresh again ;-).

gyst:
I've put this on hold so I can review with @cornae fist.

gyst:
I also found some further corner case that I'm still working on:

- scrollable is offset but not positioned
- scrollable is offset but not positioned and has overflow:auto but is not overflowing

gyst:
@cornae Can we schedule a call to review this? 

Previously, pat-scroll only properly supported scrolling the document body. In #465 I introduced a change that added support for calculating the target position in inline scrollables, which fixed a very specific scenario but is incorrect outside of that specific scenario and even caused regressions in situations that used to work correctly before.

This PR introduces a completely revamped way of calculating target offsets, if the target is contained within a scrollable that is not the document body. This requires measuring the position offset of the target versus the (possibly overflowed, clipped) top/left of the scrollable. To do that requires a positioning context to be available that represents the scrollable top/left - and we cannot use the scrollable itself since *the top of the scrollable itself does not coincide with the top of the scrollable content* except in the special case where no scrolling has been performed yet (nor does the scrollable itself necessarily provide a positioning context). Hence, I wrap the scrollable content temporarily in a helper div that provides such positioning context (by virtue of `position:relative`), perform the offset measurement, and then unwrap the scrollable content again and remove the temporary helper div before actually executing the scroll.

For the simple case that the scrollable is the document body, the complex calculation is bypassed and no helper div is injected/removed.

Additionally, I've also added support for `overflow:scroll` containers, for horizontal scrolling, for nested positioning contexts, and removed bugs in edge conditions (e.g. when a `overflow:auto` container has no scroll bar).

@pilz will be very happy when we get this merged soon.

pilz:
Yes, I would be very happy… :)

cornae:
Yes, tomorrow?

gyst:
Wednesday I'm available 9:00-10:00 and 13:30-17:30.

cornae:
9 works for me. Afternoon also if I get a specific time.

Op di 30 aug. 2016 om 21:51 schreef Guido Stevens <notifications@github.com>

gyst:
Let's do it at nine.

--
   Guido Stevens  |  +31.43.3618933  |  http://cosent.nl

   s o c i a l   k n o w l e d g e   t e c h n o l o g y

cornae:
That's a bit late notice ;). I can make it 9;30 now.
Op wo 31 aug. 2016 om 08:05 schreef Guido Stevens <notifications@github.com>

> Let's do it at nine.
>
> --
> Guido Stevens | +31.43.3618933 | http://cosent.nl
>
> s o c i a l k n o w l e d g e t e c h n o l o g y
>
>
> > On 30 aug. 2016, at 21:58, Cornelis G. A. Kolbach <

gyst:
Reimplemented based on call with @cornae .

pilz:
Seems your code got a little shorter… 💯

PR #471, opened 2016-10-01
Only look for the hashmark. 

pilz:
Until now this assumed that a url ends with a slash.

This surfaced when updating jekyll and the default to create pages in directories changed to create pages with .html. That killed the detection of which elements to clear so that all items were always cleared.


pilz:
not ready yet

PR #472, opened 2016-10-01
Pat scroll fix

pilz:
Until now this assumed that a url ends with a slash.

This surfaced when updating jekyll and the default to create pages in directories changed to create pages with .html. That killed the detection of which elements to clear so that all items were always cleared.

### DIFF
--- a/src/pat/scroll/scroll.js
+++ b/src/pat/scroll/scroll.js
@@ -61,12 +61,14 @@ define([
         clearIfHidden: function(ev) {
             var active_target = '#' + window.location.hash.substr(1),
                 $active_target = $(active_target),
-                target = this.$el[0].href.split('/').pop();
+                target = '#' + this.$el[0].href.split('#').pop();
+                log.warn($active_target);
+                log.warn(target);
             if ($active_target.length > 0) {
                 if (active_target != target) {
                     // if the element does not match the one listed in the url #,
                     // clear the current class from it.
-                    var $target = $(this.$el[0].href.split('/').pop());
+                    var $target = $('#' + this.$el[0].href.split('#').pop());
                     $target.removeClass("current");
                     this.$el.removeClass("current");
                 }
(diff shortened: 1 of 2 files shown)


src/pat/scroll/scroll.js
```diff
@@ -61,12 +61,12 @@ define([
         clearIfHidden: function(ev) {
             var active_target = '#' + window.location.hash.substr(1),
                 $active_target = $(active_target),
-                target = this.$el[0].href.split('/').pop();
+                target = '#' + this.$el[0].href.split('#').pop();
             if ($active_target.length > 0) {
```
jcbrand: I think this only works for relative URLs and will break for absolute URLs, no?

pilz: Ok, here is what I found. Let's assume that the href is pointing to a directory:
```
   target = 'http://something.de/a/b/#whatever'
=> 'http://something.de/a/b/#whatever'
   target.split('/')
=> [ 'http:', '', 'something.de', 'a', 'b', '#whatever' ]
   target.split('/').pop()
=> '#whatever'
```
Now it makes sense to do a comparison between active_target and target (line 66, not shown above):

`if (active_target != target) {`

But if the href is pointing to a file:

```
   target = 'http://something.de/a/b/c#whatever'
=> 'http://something.de/a/b/c#whatever'
   target.split('/')
=> [ 'http:', '', 'something.de', 'a', 'b', 'c#whatever' ]
   target.split('/').pop()
=> 'c#whatever'
```

then the outcome always contains the file before the hash, if I split by '/'. Comparison between active_target and target will never make sense and the if in line 66 will always be true (!=), thus the class will always be cleared everywhere - resulting in the misbehaviour.

As this target/active_target is never used for linking, only for class setting, this should be fine I think.

jcbrand: Thanks for the explanation, makes sense to me now.

pilz: Thank you. I take that as a go :)

PR #475, opened 2016-10-08
added sticky pattern

pilz:
Attempt to integrate pat-sticky. I don't know how to test if it behaves properly though and integrated the default demo, which is probably way too heavy. @cornae would you take a look and potentially reduce the demo to a viable minimum?


pilz:
@cornae should we merge?

PR #479, opened 2016-11-01
An autoload-visible element that is set hidden should not fire

pilz:


pilz:
I merge this as cornelis confirms that it is the right way. More judgement needs to come from field testing.

PR #484, opened 2017-02-14
Add a safety border to avoid modals to expand over the full height

pilz:
Attempt to get rid of the calculation issue where the modal is "just" too big for the modal.

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -19,6 +19,7 @@
 - An href can also contain a url left of the hashmark. pat-scroll should only care for the part right of the hashmark
 - Issue a delayed redraw of the calendar to prevent rendering race conditions
 - make list of calendar categories unique to speed up js processing on sites with many calendars.
+- Add a safety border to avoid modals to expand over the full height 
 
 ## 2.0.14 - Aug. 15, 2016
 
--- a/src/pat/modal/modal.js
+++ b/src/pat/modal/modal.js
@@ -121,12 +121,13 @@ define([
         resize: function() {
             var modal_height = this.$el.outerHeight(true);
             var modal_padding = modal_height - this.$el.outerHeight();
-            var max_height = $(window).innerHeight() - modal_padding;
+            // Introducing an arbitrary amount of padding that should always be given for a modal to keep a safety distance to the edges of a browser
+            var max_height = $(window).innerHeight() - modal_padding - 100;
             var $tallest_child = this.getTallestChild();
             var tallest_child_height = $tallest_child.outerHeight(true);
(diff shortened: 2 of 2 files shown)


cornae:
A safety margin must be and is handled by the CSS and may differ per design, device and screen size. If that doesn't function, then the height is probably not calculated well and then that should be fixed, not like this.

cornae:
Various improvements have been made to the Modal CSS in projects from which we should probably harvest.

pilz:
This is star. How do you think we may profit from css improvements?
The problem I am having is that the calculation compares the biggest element within the modal, adds the header size and compares it to the height of the window. I am apparently missing some pixels because that can still result in the window being bigger than the biggest element and thus the modal getting too small a dimension to fit the content

cornae:
Most of that is there for IE 8 and even 6. Today it would suffice to draw
the modal with it's content, when the content and images and all that are
there measure the modal content div, if modal content DIV's height exceeds
viewport height put class 'max-height' (we have that already) and then put
a class modal-ready on the modal root element or the body. The whole
routine is repeated when layout modifiers are triggered within the modal
and upon window resize. With a version like that I could further simplify
the CSS. I've seen miscalculations happen now and then in Star too, but I
think it has to do with the current height calc script.

Op wo 15 feb. 2017 om 14:19 schreef Alexander Pilz <notifications@github.com

pilz:
The problem seems to be that the height of the pat-modal div here is returned as 925, the window inner height is returned as 1028, so all seems fine, but still the contents of the modal are too large to accomodate:

bildschirmfoto 2017-02-15 um 14 41 08

cornae:
Maybe it measures the pat-modal DIV at a moment when there's max-height on
it which fixes the modal? Also, currently I would rather measure
DIV.panel-content than DIV.pat-modal. Seems far less error prone to me as
you then stay away from possible race conditions.

Op wo 15 feb. 2017 om 14:44 schreef Alexander Pilz <notifications@github.com

pilz:
Yes, seems something like that. I get accurate readings if I measure panel-content and add header. That gives me the total size that the browser then reports for pat-modal in total.

cornae:
I would try then to just look at the height of panel-content and check if
that exceeds the content height (inner height) of .pat-modal and in that
case place the max-height class. All other operations such as setting
height attributes become likely obsolete. Some CSS updates are required to
make it work.

Op wo 15 feb. 2017 om 14:51 schreef Alexander Pilz <notifications@github.com

pilz:
I am closing this in favor of #487, which solves the problem by simplification

PR #485, opened 2017-02-14
Don't close a tooltip if validation warnings show

pilz:
@jcbrand do you have perhaps an idea why this works and whether it poses any problems?

In calendar, when I add an event by clicking into a day, a tooltip with the add event form opens. If I enter an invalid date or don't enter a title, validation is triggered, throws a pat-update, catches that itself, returns true and the tooltip closes. 

If I catch it like in this PR, all seems to behave fine.

### DIFF
--- a/src/pat/validation/validation.js
+++ b/src/pat/validation/validation.js
@@ -288,6 +288,10 @@ define([
                 if (data.pattern == "clone" && data.action == "remove") {
                     this.validateForm(ev);
                 }
+            } else if (data.pattern == "validation") {
+                // validation throws pat-update itself. If that happens within a tooltip, the tooltip closes.
+                // returning false here prevents this. @jc, any idea why this is a fix and whether this is harmful?
+                return false;
             }
             return true;
         },


jcbrand:
From the jQuery docs:
```Returning false from an event handler will automatically call event.stopPropagation() and event.preventDefault().```

So by returning `false` you're making sure that `pat-update` doesn't propagate further (eventually causing the tooltip to close).

There might be side-effects due to this change. Ideally one would figure out how `pat-update` causes the tooltip to close and to work against that inside the tooltip pattern itself.

It might be enough to have a similar event handler that returns false, but in the tooltip pattern and not in pat-validate.

pilz:
Thank you. I am trying to identify where that close happens, and it is actually the destroy method called on pat-tooltip that closes it. However this is not called from within pat-tooltip, so now I am hunting for anything that might be triggered by pat-update and results in a destroy event propagated to pat-tooltip.

Is that only me or is the js event model quite intransparent to debug?

jcbrand:
Events can be difficult to debug because the call stack can end abruptly before reaching the line where the event is fired. 

That said, if you haven't done so yet, look at the call stack in the developer tools. Sometimes you're able to go up all the way to the line of code that triggered the event.

pilz:
This is actually invalid. Closing in favor of #486

PR #486, opened 2017-02-15
make calendar respect the new validation pattern

pilz:


pilz:
Thank you @jcbrand, the info about the call stack made me realise this!

PR #487, opened 2017-02-15
simplify the calculation by checking if the content is larger than th…

pilz:
…e calculated modal div and set the max-height class then.

### DIFF
--- a/src/pat/modal/modal.js
+++ b/src/pat/modal/modal.js
@@ -119,21 +119,14 @@ define([
         },
 
         resize: function() {
-            var modal_height = this.$el.outerHeight(true);
-            var modal_padding = modal_height - this.$el.outerHeight();
-            var max_height = $(window).innerHeight() - modal_padding;
-            var $tallest_child = this.getTallestChild();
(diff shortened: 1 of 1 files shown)


src/pat/modal/modal.js
```diff
@@ -119,22 +119,15 @@ define([
...
+            var header_elem = this.$el.find('.header');
 
-            if (tallest_child_height !== modal_height) {
-                modal_height = tallest_child_height + header_size + modal_padding;
-            }
-            if (max_height < modal_height) {
+            var modal_height = panel_content_elem.outerHeight(true) + header_elem.outerHeight(true);
+
+            if (this.$el.height() < modal_height) {
                 this.$el.addClass("max-height").css("height", max_height);
```
ale-rt: @pilz we have an error here that prevents the display of the modal.

```
patterns.pat.modal: Failed while initializing 'modal' pattern. ReferenceError: max_height is not defined
```

The fix is either to set `max_height` to a proper value or '100%' . I am making a PR for this.

PR #488, opened 2017-02-20
Fix undefined variable max_height

ale-rt:


ale-rt:
@pilz this fixes the undefined variable problem. Anyway I am still unsatisfied because there are rounding problems here:

```
            var modal_height = panel_content_elem.outerHeight(true) + header_elem.outerHeight(true);
            if (this.$el.height() < modal_height) {
```

sometimes this.$el.height() - modal_heigth = -1.

This makes the scrollbar appear also when it shouldn't.

Probably we should change the condition like this:

```
if (modal_height - this.$el.height() > 1) { 
```

ale-rt:
An attempt to use that condition and to simplify the code has been done here:

- https://github.com/ale-rt/Patterns/commit/fc704c8d8dc469163b0c53a76a4f174b2dc19736

pilz:
agreed. Sounds good

ale-rt:
I would say let's merge this one that fixes the undefined variable.
Then I can test locally for a while this patch:

- https://github.com/ale-rt/Patterns/commit/fc704c8d8dc469163b0c53a76a4f174b2dc19736

before making a new PR

PR #491, opened 2017-02-27
pat-masonry: initialize just before layouting

thet:
pat-masonry: Initialize masonry just before layouting gets startet, which is after first image has been loaded or at loading has finished.
This avoids overlapping images while they are still being loaded.


thet:
@pilz @jcbrand please review

PR #492, opened 2017-03-01
Bump modernizr

pilz:


pilz:
Cornelis wants/needs always the newest version

pilz:
@cornae this branch upgrades modernizr to the newest version. This also has an effect on pat-bumper, where we - back then - had to manually include the csspositionsticky check which is now included in modernizr by default. As you are also looking at pat-sticky and other applications of the bumper pattern, you might want to recheck if the old pat-bumper now behaves better than before.

pilz:
I am seeing problems with classes not showing on the html tag. Investigating.

pilz:
Again, things are not that easy. Modernizr now comes configurable, like patternslib does. So you can specify which detects are included. There are currently 259 detects, resulting in a 303k large file. Massive. And we can't simply download modernizr anymore, it needs to be built. Bower can't do that so we need our own process for this - or generate the file manually and check it in - and update once in a while. Meh.

jcbrand:
@pilz It might be that the dist files are included when you install it via NPM and not Bower because for NPM you'll always get releases, not just a checkout of the source code.

You can set the version to `"*"` to get the latest version.

There's no longer any technical reason to use Bower and it's often inferior to using NPM. I moved all my bower dependencies for converse.js over to NPM.

pilz:
This is no longer necessary. The upcoming update to webpack and npm also comes with the latest modernizr. See branch deprecate-bower for details.

PR #494, opened 2017-03-04
Add colour-picker to demo page

janga1997:


janga1997:
@pilz This seems like an error not related to my PR. Could you help out ?

janga1997:
That's weird. I just triggered the tests with an empty commit, and it seemed to pass.

pilz:
Sometimes, travis also just has some issues with the network or doesn't get a dependency correctly. What you did is just right, if the error doesn't seem related to your change, just give it a second try :)

PR #496, opened 2017-03-06
documentation and tests for menu

janga1997:
Documentation for menu pattern, and add `Chosen` to demo page.
I would also like to get started with writing tests ( for some patterns in `Forms` ), so it would be great if anyone had any tips.

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,7 +24,7 @@
 			<a class="navigation tree" href="/src/pat/expandable-tree/index.html" title="">Expandable tree<sup>unknown</sup></a>
 			<a class="navigation navigation-marker" href="/src/pat/navigation/index.html" title="Navigation markers">Navigation<sup>unknown</sup></a>
 		</div>
-		
+
 		<div class="layout">
 			<h4>Layout</h4>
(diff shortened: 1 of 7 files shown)


jcbrand:
Type `make check` to run the tests.

The tests use Jasmine 1.3 and can either run in the browser or in the console via phantomjs (which is how they run when you type `make check`).

If you want to run them in the browser, you should type `make serve`, and then load `localhost:4001/tests.html` in your browser.

The tests for each pattern goes into the pattern's own directory. They are then loaded for testing in this file: https://github.com/Patternslib/Patterns/blob/master/tests.js

janga1997:
Thanks for the info.@jcbrand  !
Although I meant tips on writing the actual tests, which I'm finding little hard to begin with.

janga1997:
I'm facing a bit of trouble in writing the test. 'Pattern' seems to be always undefined in the statement `Pattern.init`, even though it is, and I followed the template of other tests.

thet:
Closing this due to follow up: https://github.com/Patternslib/Patterns/pull/696

PR #497, opened 2017-04-06
Also reload page when browser back is fired

pilz:


reinhardt:
I noticed that there's already a window.location.reload() in the code and that it would do the same, however the condition is never met. I've removed the condition; that makes the self-assignment obsolete.

PR #500, opened 2017-04-20
Fix nested autosubmit

ale-rt:


cornae:
Wow this sounds like heavy risky stuff. Would it be worth to have a call
and talk use cases through the way we always do when inject or auto submit
or subform or touched?
Op do 20 apr. 2017 om 15:36 schreef Alexander Pilz <notifications@github.com

Issue #501, opened 2017-04-25
New release?


agitator:
last one is 2.0.14 - Aug 15, 2016

cornae:
I think it's soon a good time for a new release too. There have been several improvements lately.

pilz:
I took over from JC and did a 2.1.0 following the RELEASE.md notes. Please have a look and see if it is properly usable. Let me know of any issues with the release process.

PR #503, opened 2017-05-23
gallery: allow pat-gallery trigger on image wrapper directly

thet:
pat-gallery: Also include the node with the ``pat-gallery`` class trigger for initializing the gallery.
Now ``pat-gallery`` can be used on anchor tags wrapping images directly, making it possible to let images be opened individually in in the overlay without adding them to a gallery with navigation controls to the next image.


cornae:
Sounds great. Is the documentation also updated?

thet:
@cornae tnx for the pointer!
I updated the PR with docs updates including info about the ``item-selector`` property, which I added in a previous PR which is already merged.

cornae:
The property item-selector is not mentioned in the documentation on the site either: http://patternslib.com/gallery/  I'm super happy about knowing this because I posted a ticket in a project only a week ago with a request for exactly this behaviour. Very happy it's in :).

thet:
tnx for merging!
is the documentation on http://patternslib.com/gallery/ in a separate repository?

pysailor:
Hi @thet,
no, the documentation is fine where you put it. But http://patternslib.com/gallery/ does not update automatically.
I just updated the site, now your new documentation is there.

(Note that downloading the bundle from the site seems to be broken ATM, we'll need to fix that, also in combination with the new release #501)

PR #506, opened 2017-06-08
Resize a masonry that contains an equalize after the latter has been …

ale-rt:
…updated

This is a follow up of #505.

If a .pat-masonry element contains a .pat-equalise, when this one is updated it will trigger a `pat-updated` event that will cause the parent pat-masonry to resize properly and will also resize its children pat-masonry.


ale-rt:
@pilz Can you check this one please?

PR #516, opened 2017-09-01
Deprecate bower

pilz:
This is a rather large PR which switches from require.js to webpack and deprecates bower. It also updates a few libraries and upgrades tests to jasmine 2.8 and uses karma to run them. 
This is the preparation for a larger scale maintenance offensive that we want to undertake.

As this is a breaking update, I upgrade the version to 3.

There is a file "Upgrading-from-2-to-3.txt" in the root to explain the steps. 

As this might break for users who use master, I would like to hear comments about merging. If I don't hear anything, I could just merge. If comments are cautious, we might go for a more defensive approach to carry on with separate branches for a while. 

Let me know :)


pilz:
@jcbrand @thet @wichert @vangheem et al, any opinions?

wichert:
I certainly like the initiative. A next step would be to move to ES6 imports as well - I'm actually surprise the require() approach works with webpack.

pilz:
Yes, ES6 should be possible now with that. The fact that require worked was a big plus for this.

jcbrand:
@pilz How does this impact the bundler on the Patternslib website, now that require.js is no longer used?

pilz:
@jcbrand As discussed, would you merge this?

jcbrand:
@pilz Done, sorry for the wait.

pilz:
Thank you!

thet:
+1 for the change.

By the way, for mockup, we're planning a move to npm and a restructuring too. See: https://github.com/plone/Products.CMFPlone/issues/1653

I'm tempted to start right implementing PLIP 1653, but unfortunately I have some other work to get done.

PR #518, opened 2017-09-14
vip renames: rename upgrade guide and changes to meet conventions.

thet:
super important renames.

i've just done it - the upgrade guide's filename was wrong in changes.md, also it was a .txt file with MarkDown contents, then i saw all other doc file names written in uppercase, simplified the name, ....


pilz:
Perfect, thanks :)

PR #524, opened 2017-09-21
Update dependencies

ale-rt:


pilz:
I am not sure this is getting us into calmer water :) I picked the latest fullcalendar 2 version because 3 introduces severe changes that would require another design round. And we are using some code in patterns that is known not to work with latest jquery. So these pinnings will need more testing

ale-rt:
This was just explorative.

pilz:
Boy, you scared me there :)

PR #525, opened 2017-09-22
Wait for injection to be triggered before closing the tooltip

ale-rt:
On recent chrome versions, if you have tooltip containing a `form.pat-inject`, the form may not be submitted when you submit it clickiing on a `.close-panel` button with the following error:
```
Form submission canceled because the form is not connected
```

This patch prevents the error hiding the tooltip only after the injection is triggered.

Closes ploneintranet/issues#1498


pilz:
Great!

PR #529, opened 2017-10-23
Upgrade sizzle

pilz:


jcbrand:
Sizzle is usually pulled in as a dependency of jQuery. I would therefore be careful to update it individually while keeping jQuery as is. Perhaps consider updating jQuery as well?

pilz:
Yes, that is the case, so pinning doesn't make a difference. I remember that updating jQuery may have implications on the parser, so perhaps I'll try a quick bump just to see the extent of the damage, but this is probably not a quick solution. Currently I see failue in masonry which doesn't render in IE11 due to a sizzle bug. I'll try updating masonry first.

PR #538, opened 2017-11-23
New carousel

pilz:
Based on slick (http://kenwheeler.github.io/slick/)


Makefile
```diff
@@ -74,7 +74,7 @@ all_css:: css
 	@$(SASS) -I . -I _sass src/pat/bumper/_bumper.scss src/pat/bumper/bumper.css
 	@$(SASS) -I . -I _sass src/pat/calendar/_calendar.scss src/pat/calendar/calendar.css
 	@$(SASS) -I . -I _sass src/pat/carousel/_carousel.scss src/pat/carousel/carousel.css
-	@$(SASS) -I . -I _sass src/pat/checked-flag/_checked-flag.scss src/pat/checked-flag/checked-flag.css
+	@$(SASS) -I . -I _sass src/pat/carousel/_carousel-legacy.scss src/pat/carousel-legacy/carousel-legacy.css
 	@$(SASS) -I . -I _sass src/pat/checklist/_checklist.scss src/pat/checklist/checklist.css
```
wichert: Did you mean to remove checked-flag?

webpack/base.config.js
```diff
@@ -124,6 +125,7 @@ module.exports = {
             "pat-bumper": "pat/bumper/bumper.js",
             "pat-calendar": "pat/calendar/calendar.js",
             "pat-carousel": "pat/carousel/carousel.js",
+            "pat-carousel-legacy": "pat/carousel-legacy/carousel-legacy.js",
```
wichert: Perhaps remove this from the default build?

cornae:
@wichert, The likelyhood  likely that any existing deployments of pat-carousel will break is 100%. The likelihood that the old pat-carousel was ever used is something I cannot measure but I expect it's close to 0, because it was never a rather cumbersome script and it didn't have touch support. The alternative is to call the script pat-slick. The advantages are that it doesn't break legacy patterns carousels (I don't have any, but they might exist in the world) and that it gives up front credit to the slick library. The disadvantage is that 'slick' as a name doesn't give away what (interaction) pattern it is. So all balanced out, pat-carousel seems a fine choice with me.

PR #539, opened 2017-11-27
Fix pat-switch

ale-rt:
This fixes the injection of links with the option `trigger: autoload-visible` inside an hidden element made visible after a pat-switch event.

### DIFF
--- a/src/pat/switch/switch.js
+++ b/src/pat/switch/switch.js
@@ -83,6 +83,7 @@ define([
                 if (option._storage)
                     option._storage.set(option.selector, {remove: option.remove, add: option.add});
             }
+            $trigger.trigger('resize');
         },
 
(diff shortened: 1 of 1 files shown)


ale-rt:
@pilz maybe you know a better fix for this one: triggering a resize may be too much.

ale-rt:
I put the old merge because after I will know the proper fix, I will add a more detailed changelog line.

pilz:
Apparently we have a situation where a pat-switch triggers a layout change that requires a redraw. However that is not necessarily always the case for a pat-switch. However that lies in the css domain and not in the js domain, so we can't know in advance. At this moment it appears sane to me to assume that a layout relevant change may happen and that it is important to assume that a redraw is needed. Based on that rationale, I am merging this.

PR #544, opened 2017-12-04
Fix jshint

ale-rt:


ale-rt:
I think the build was broken by some recent changes in master.

PR #547, opened 2017-12-04
do explicitly test for undefined, you are not a python programmer, da…

pilz:
…mmit


jcbrand:
@pilz Patterns has underscore.js available, so you can do `_.isUndefined(variable)` or even `_.isNull(variable)` and to check for either `undefined` or `null` you can do `_.isNil(variable)`.

pilz:
Hi JC, thank you, this is a real relief :)

PR #548, opened 2017-12-06
check if the pat-update comes from pat-equalise and if so, don't upda…

pilz:
…te again

### DIFF
--- a/src/pat/equaliser/equaliser.js
+++ b/src/pat/equaliser/equaliser.js
@@ -8,8 +8,9 @@ define([
     "pat-registry",
     "pat-parser",
     "pat-utils",
-    "imagesloaded"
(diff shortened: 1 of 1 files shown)


src/pat/equaliser/equaliser.js
```diff
@@ -67,7 +68,10 @@ define([
             $container.trigger('pat-update', {pattern: 'equaliser'});
         },
 
-        _onEvent: function(event) {
+        _onEvent: function(event, extra) {
+            if (!_.isUndefined(extra) && extra.pattern === 'equaliser') {
```
ale-rt: I think we should think about a smarter condition because  this breaks nested masonry things:
image

Without this patch it works great:
image

pilz:
invalid, we'll do this with MutationObservers.

PR #552, opened 2017-12-20
Make code more solid

ale-rt:
This prevents an error when `_._findSiblings` exits returning undefined.

### DIFF
--- a/src/pat/checklist/checklist.js
+++ b/src/pat/checklist/checklist.js
@@ -72,6 +72,7 @@ define([
         onChange: function(event) {
             var $trigger = event.data.trigger,
                 options = $trigger.data("patternChecklist");
+            var siblings;
 
 
             var all_selects = $trigger.find(options.select);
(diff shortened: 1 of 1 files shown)


src/pat/checklist/checklist.js
```diff
@@ -79,15 +80,16 @@ define([
                 all_deselects = $(options.deselect);
             }
             for (var i=0; i<all_selects.length; i++) {
-
-                if (_._findSiblings(all_selects[i], "input[type=checkbox]:visible").filter(":not(:checked)").length === 0) {
+                siblings = _._findSiblings(all_selects[i], "input[type=checkbox]:visible");
+                if (siblings && siblings.filter(":not(:checked)").length === 0) {
```
pilz: Not sure if this applies here as well but I often had to test explicitly for undefined, because any other value from undefined, even if 0 or empty string would make "siblings" true. Perhaps explicit is better than implicit here. (siblings != undefined &&…)

src/pat/checklist/checklist.js
```diff
@@ -79,15 +80,16 @@ define([
...
                     $(all_selects[i]).prop("disabled", true);
                 } else {
                     $(all_selects[i]).prop("disabled", false);
                 }
             }
             for (var i=0; i< all_deselects.length; i++) {
-                if (_._findSiblings(all_deselects[i], "input[type=checkbox]:visible").filter(":checked").length === 0) {
+                siblings = _._findSiblings(all_deselects[i], "input[type=checkbox]:visible");
+                if (siblings && siblings.filter(":checked").length === 0) {
```
pilz: same as above

src/pat/checklist/checklist.js
ale-rt: Are you sure that 0 or '' will evaluate to `true`?
```
[ale@emily Patterns]$ node
> if (0) {console.log('I am true')};
undefined
> if (1) {console.log('I am true')};
I am true
undefined
```

pilz: Ok, my case was different (https://github.com/Patternslib/Patterns/pull/547/files), here I got a jQuery object back. So this is good :)

ale-rt: BTW that condition is really strange... a || !a && b should equal to a || b. It has a bug, it should be simplified or I am reading it wrong.

PR #554, opened 2017-12-20
Clicking on an autoload visible link should inject properly

ale-rt:
Fixes an issue that happens when the autoload visible link is injected in a collapsed element (see #555).
The collapsed element has no scrollbar until it is expanded.
When it is expanded and we scroll down until the link is visible it is already too late: the link will not be injected.
Clicking it was not triggering the injection.
Only resizing the window triggered the injection.


ale-rt:
Although #555 should prevent this problem to happen, I would say it is anyway good to have this merged in case other edge cases will prevent the link to properly auto inject itself when it becomes visible.

PR #557, opened 2017-12-21
Fix double click on an injection link

ale-rt:


ale-rt:
Maybe something similar is needed also in some other cases (e.g. form submission)

PR #567, opened 2018-01-03
Support time units

wichert:
This PR add a utility method to parse time periods. It assumes milliseconds by default (which guarantees backwards compatibility). Supported units are:

* `s`: seconds
* `ms`: milliseconds
* `m`: minutes

This is currently used by the delay option for the injection and tooltip patterns.


cornae:
Super!

PR #569, opened 2018-01-05
Mutationobserver masonry

pilz:
Don't merge yet, this is a request for comment for now.

Attempt to implement Wichert's suggestion of mutation observers. I am looking for hints of what side effects I might have missed and might need to take into account.

We also discussed to make masonry rendering less expensive by storing if we need to recalculate at all. That is not in here yet.

### DIFF
--- a/src/pat/equaliser/equaliser.js
+++ b/src/pat/equaliser/equaliser.js
@@ -24,13 +24,24 @@ define([
                 var $container = $(this),
                     options = parser.parse($container, opts);
                 $container.data("pat-equaliser", options);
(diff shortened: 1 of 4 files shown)


pilz:
Just tested with a manual build on star, it does fix the issue where the stream doesn't update when the sidebar is expanded.

cornae:
That sounds promising. I would like to test it against many more cases in a complex product like Quaive. Is that possible?

pilz:
Yes, I am going to mail you a ploneintranet.js to manually drop in for tests.

pilz:
No comments and no problems found. So I will merge it now.

PR #571, opened 2018-01-10
bump to 4.2.2 to fix .ajaxSubmit not a fucntion in form injections

pilz:
@wichert Cornelis reports this is broken since before christmas, but I didnt identify what broke it. 
Submitting a form through injection resulted in the following error

bildschirmfoto 2018-01-10 um 14 53 05


cornae:
I can still reproduce the error though.

PR #573, opened 2018-01-11
Replicate PR #564 for pat-checked-flag

ale-rt:
See #564

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,7 +48,7 @@ Breaking Changes
 - Extend pat-focus to add `has-value` class, `data-placeholder` and `data-value` attributes.
 - Add idle trigger to injection.
 - Fixed injection so that urls with data: in them don't get prefixed with a / anymore.
-- pat-checklist now uses Sets to collect its siblings, that should make it much faster with large lists of icons.
+- pat-checklist and pat-checked-flag use Sets to collect siblings, that makes them much faster with large lists of icons.
 - pat-carousel gets infinite option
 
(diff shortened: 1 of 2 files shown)


pilz:
Actually we shouldn't use pat-checked-flag anymore, it should be superseeded by pat-checklist. Where did you find it? Perhaps we can just replace it?

cornae:
Correct.

ale-rt:
I found it in portal stardemo. I think that pattern is triggered too easily ;) https://github.com/Patternslib/Patterns/blob/master/src/pat/checked-flag/checked-flag.js#L19

cornae:
It must die. :)

ale-rt:
Should I just remove it?

pilz:
I think we should not remove it yet. Perhaps better output a warning and do another grep where it might be used. I guess I only grepped in patternslib but not in the projects, as you found this occurrence. If you just remove the pattern, things might break silently.

ale-rt:
I think this one is silently triggered because of this line: 
```
trigger: "input[type=checkbox],input[type=radio],select",
```

pilz:
Aaaah, in this case, it would be enough to remove the pat-checked-flag pattern from the star patterns.js, so that it doesn't get included anymore. I think that is, what you meant in your comment above, right? If so, please do.

pilz:
I merge this as a last compatibility fix for anybody who still uses it but a note to anybody who might read this: checked-flag is deprecated and should be replaces by selectbox.

PR #520, opened 2017-09-15
New pattern: datetimepicker, Masonry enhancements

thet:
This targets the 2.1.x branch. A forward port to master is at: https://github.com/Patternslib/Patterns/pull/575

## pat-masonry
 - Re-Layout on ``load`` events emitted on ``img`` nodes within the masonry scope.
 - Do not depend on ``imagesloaded``.
 - Update masonry to version 4.2.0.
 - Align options with new version:
   - Add options ``is-horizontal-order``, ``is-percent-position`` and ``is-resize``.
   - Remove options ``visible-style`` and ``hidden-style``.
   - Add Aliases from v4 (no ``is-*``) to v3 names (with ``is-`` for booleans), while keeping the v3 names. The patternslib parser does boolean casting for ``is-*`` options.

## pat-datetime-picker
- Add new pattern for setting the date and time.

## pat-date-picker
- Remove the dependency on ``moment-timezone-data`` - it's not used and there is no use in a date picker anyways.

## Makefile
- with the ``clean`` target, also remove the ``stamp-bundler`` marker file.

PR #575, opened 2018-01-26
 New pattern: datetimepicker, Masonry enhancements

thet:
This is the forward-ported version of https://github.com/Patternslib/Patterns/pull/520

## pat-masonry
 - Re-Layout on ``load`` events emitted on ``img`` nodes within the masonry scope.
 - Do not depend on ``imagesloaded``.
 - Update masonry to version 4.2.0.
 - Align options with new version:
   - Add options ``is-horizontal-order``, ``is-percent-position`` and ``is-resize``.
   - Remove options ``visible-style`` and ``hidden-style``.
   - Add Aliases from v4 (no ``is-*``) to v3 names (with ``is-`` for booleans), while keeping the v3 names. The patternslib parser does boolean casting for ``is-*`` options.

## pat-datetime-picker
- Add new pattern for setting the date and time.

## pat-date-picker
- Remove the dependency on ``moment-timezone-data`` - it's not used and there is no use in a date picker anyways.


thet:
A @pilz how can I run the tests? ``make check`` apparently doesn't run any tests on my computer.

ale-rt:
HI @thet, tests are broken also on master, see #579.
I am restoring them #578, see.
After #578 is merged rebase to master and run `make check`.

If you want to run the test locally on Chrome run:
```
karma start webpack/karma.conf.js --watch --browsers Chrome
```
and go to http://localhost:9876/debug.html.
It should work after `make check` prepares all the needed stuff.

jensens:
@pilz I tried to rebase this PR, but I don't have permissions on the repository. May I ask kindly to get permissions to contribute here?

pysailor:
@jensens I just sent you an invite for Patternslib, can you please confirm it an re-check your access?

jensens:
@pysailor thanks, access works!

jensens:
phew, seems this PR should be split up in two. One with the masonry, the other one with the datepicker stuff.

thet:
Green as green jelly beans. Nice.

jensens:
So, this one is ready too!

pilz:
:)

src/pat/masonry/masonry.js
```diff
@@ -12,51 +12,54 @@
...
-                    this.initMasonry();
-                }
-                this.quicklayout();
-            }.bind(this));
-            imgLoad.on("always", function () {
-                if (! this.msnry) {
-                    this.initMasonry();
-                }
-                this.layout();
-            }.bind(this));
```
pilz: Hi @thet, can you tell me how you replace the imagesloaded functionality? This commit looks like you basically replaced that with the initMasonry call only. But the layout call is missing, so my masonries are not rendering. How are you doing this?

pilz: I don't yet understand how the imagesloaded part is being replaced. I see timing issues again where the masonry is layouting while not all images are loaded yet. That results in views like the attached screenshot. Until we find out, I am tempted to readd the imagesloaded code.
bildschirmfoto 2018-06-17 um 16 56 48

thet: The replacement for imagesloaded is in the following lines in the 
https://github.com/Patternslib/Patterns/pull/520 PR for the 2.1.x branch. Especially the "load" event handler on image elements:
```
            // Initially layout on document ready.
            $(document).ready(utils.debounce(this.layout.bind(this), 100));

            // Re-Layout, if images are loaded within pat-masonry
            $('img', this.$el)
.on("load", utils.debounce(this.quicklayout.bind(this), 200));
```

That got lost in this PR. Here an observer is used.

thet: @pilz does this look only when the site is loaded or even after all images have been loaded?

pilz: The observer is looking for changes in the DOM only. That one addresses issues when something is injected outside the masonry context so that masonry can adapt to the new space situation and rerender. images loaded is not addressed by the observer. Can you give me some background why the img.load event handler is better? Is that something new? Why does imagesLoaded exist if you can do this that easy? And did imagesLoaded give you some problems so that you had to replace it with the on load handler?

thet: @pilz I had some issues with image loading and masonry resizing. It did look quite ugly - all images were displayed on top of each other until they were finally loaded. The optimization fixed that somehow. I saw the ``load`` event making imagesloaded obsolete - having fewer dependencies is also a nice thing.
I remember to have seen some docs hinting that the load event finally works in modern browsers and does not suffer from the issues mentioned here: http://api.jquery.com/load/ - but I can't find it yet and that might be a wrong memory. When I find out more, I'll let you know....

PR #580, opened 2018-01-30
Revert "Merge pull request #576 from Patternslib/ie11-compat"

### LINKED ISSUE #577: Pat-checklist with radio buttons glitch
Since the latest update of pat-checklist there's a problem with checklists that use input type=radio. When one of the radio buttons gets checked, the labels of the other input type=radio instances that have the same name keep their class 'checked'.

It can be reproduced in Star by opening the search app and click on 'Advanced search'. The label 'Advanced search' should be replaced by a label 'Standard search', but it doesn't because of the aforementioned issue.

ale-rt:
This reverts commit 6b25ebe080be9836114dc94cd50759af46104133, reversing
changes made to c0b585496e4f8559423be356590cce9ff204fc3c.

Should fix #577

### DIFF
--- a/src/pat/checklist/checklist.js
+++ b/src/pat/checklist/checklist.js
@@ -147,10 +147,7 @@ define([
             $result = $related=$related.not(el);
             $result.each(function(item) {result.add($result[item])});
             for (var i=0; i<$related.length; i++) {
-                var label_and_fieldset =  _._getLabelAndFieldset($related[i]);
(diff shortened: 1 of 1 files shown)


pysailor:
@ale-rt But this will break IE11 again?

ale-rt:
Of course!
And this will not be really tested by travis, as it is happening since some time

pysailor:
As we just discussed on the phone, this PR was the result of a misunderstanding. The IE11 fixes are great :-)

PR #581, opened 2018-01-30
Fix pat-checklist not working properly on radio inputs

### LINKED ISSUE #577: Pat-checklist with radio buttons glitch

ale-rt:
Closes #577


cornae:
Works like a charm in IE and all. Thanks for all the hard work!

PR #602, opened 2018-06-14
hotfix to get masonry to layout again until we sort out why it was re…

pilz:
…moved

### DIFF
--- a/src/pat/masonry/masonry.js
+++ b/src/pat/masonry/masonry.js
@@ -58,7 +58,7 @@
 
             // Initialize
             this.initMasonry();
-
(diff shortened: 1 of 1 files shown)


pilz:
Currently in PI proto, all the masonries don't render and we have the eternal spinner because masonry gets initialised but the layout command is never issued. Therefore the opacity is never set to 1 and the spinner doesn't go away. Calling the layout in init fixes it for now, but I don't understand that fully yet. To be able to investigate further, I add this quickfix, but will continue to amend once I have heard back from @thet for the reason of that change.

PR #603, opened 2018-06-18
Readd imagesloaded

pilz:


pilz:
This PR keeps the updated masonry but readds the imagesloaded part.

thet:
@pilz no problem with re-adding imagesloaded, but it shouldn't be necessary. See my comment at https://github.com/Patternslib/Patterns/pull/575#discussion_r195937420

The following should essentially do the same as imagesloaded at a lower cost:
```$('img', this.$el).on("load", utils.debounce(this.quicklayout.bind(this), 200));```

But this line got lost in PR #575 

http://api.jquery.com/load-event/ and https://stackoverflow.com/a/3877079/1337474 (from 2010) do mention some problems with the JavaScript load event, but I'm not sure how valid this information is, nowadays.

PR #604, opened 2018-06-29
pat-calendar: Check for existing categories, not for selected ones.

reinhardt:
This check is meant to make sure that we see all events in case there are no category controls. However, it mistakenly checks for any *selected* categories, not for any *existing* category controls. This means that if there are category controls and they are all deselected then all events are shown. This patch fixes the check.

### DIFF
--- a/src/pat/calendar/calendar.js
+++ b/src/pat/calendar/calendar.js
@@ -472,7 +472,7 @@ define([
                     log.debug("remove due to search-text="+searchText, $event);
                     return false;
                 }
-                if (shownCats.length === 0) {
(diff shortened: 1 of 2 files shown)


src/pat/calendar/calendar.js
```diff
@@ -472,7 +472,7 @@ define([
                     log.debug("remove due to search-text="+searchText, $event);
                     return false;
                 }
-                if (shownCats.length === 0) {
+                if ($el.$catControls.length === 0) {
```
pilz: Good catch. This is most certainly a relic from the calendar upgrade in Star P8

PR #619, opened 2018-08-29
demo how we could show inject errors

pilz:
This PR adds code to the error handler that takes the error and puts it into a notification at the end of the body. 
This is just a proof of concept for @cornae to validate.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "patternslib",
-  "version": "3.0.0a4",
+  "version": "3.0.0a5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


pilz:
@cornae this is a first attempt. 
Can we assume that everybody who uses patterns has notification support when using injection? Do we need to care? Even without notifications, the error is "just" added to the end of the body and at least gets visible.
I have added code and real world wording for the 404 case. More cases of errors would need to get added, but most are not testable with jekyll (E.g. I cant trigger jekyll to throw any http status code I want).
What do you think?

cornae:
Sounds exciting! Injecting an element could be a tiny intrusive because it will force one to write CSS for that part before an update to a newer Patterns version is possible. Would an attribute like data-error-message with the error message as its value? Then it's turned around because that value can be made visible with CSS, but will remain invisible when the user of Patterns doesn't do anything.

cornae:
schermafbeelding 2018-09-03 om 20 42 05

schermafbeelding 2018-09-03 om 21 08 04

schermafbeelding 2018-09-03 om 21 09 37

I've made several designs for the error message. The third of the ones depicted above is now in proto. It will appear you'll have to set a data attribute on body: `data-error-message="Your message"`.

Running example: http://quaive.syslab.com/workspaces-error

PR #620, opened 2018-08-31
Attempt to prevent double submits

pilz:
ported from mockup

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "patternslib",
-  "version": "3.0.0a4",
+  "version": "3.0.0a5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


jcbrand:
@pilz Without know too much about how this change works... you might want to take a look at the [debounce](https://lodash.com/docs/4.17.10#debounce) method from lodash/underscore.

You can use it to collapse multiple calls (within a short time period) to a function into a single call that executes at the beginning or end of the sequence of calls.

Sounds like it might be handy here.

pilz:
Oh, I am just porting the mockup code back to patterns, haven't really looked into the details yet. But thank you, I will keep that prominently in my mind!

cornae:
Why is this a separate pattern?

pilz:
I am still exploring. Nothing official yet. But due to our big deployment chain, I need a vehicle. I will prepare a proposal once I know what options I have.

cornae:
A lot could be prevented with CSS also when a class ‘executed’ or similar would be set on the element that has pat-inject in it while that injection is being executed. Even with user feedback + double execution prevention.

pilz:
We have some users who always double click and that results in double case creation. And as cases come from templates already with content in it, this is an operation that is heavy, then results in conflicts and we have to manually clean up. I am more cleaning up than fixing at the moment, so I am in urgent need of a quick fix for the create workspace panel. I am using this pattern to investigate how I at all can catch the click event.

cornae:
I fully understand the case and the urgency of it. Even if the ticket has laid still for some time. My ideas for a the lightweight and easy fix are still here…

pilz:
Can you help quickly with a fix? I need something to ease the situation so that I can clean-headedly rethink the whole complex.

cornae:
Sure. I'm available for a call.

pilz:
Thank you, that was very helpful. I will revert that change.

PR #625, opened 2018-10-16
pat-inject: autoload-visible now uses the intersection observer

ale-rt:


ale-rt:
Skip whitespaces in the diff with https://github.com/Patternslib/Patterns/pull/625/files?w=1

src/pat/inject/inject.js
```diff
@@ -818,21 +818,34 @@ define([
...
+                        return false;
+                    }
                     if (!utils.elementInViewport($el[0])) {
                         return false;
                     }
                     // check if the target element still exists. Otherwise halt and catch fire
-                    var target = $el.data("pat-inject")[0].target.replace(/::element/, '');
-                    if ($(target).length === 0) {
-                        return false;                        
+                    var target = ($el.data("pat-inject")[0].target || cfgs[0].selector).replace(/::element/, '');
```
ale-rt: @pilz This fixes the missing target definition in data-pat-inject, still I am not sure if this is complete because we can have multiple injection targets, right?

ale-rt:
@pilz I updated the PR with 7ac208f to also fix autoload-visible  inside a scrollable.
This will fix the issue you can see scrolling to the bottom this page: http://quaive.syslab.com/dashboard.html

Issue #629, opened 2018-10-25
pat-date-picker: i18n not working any more


pysailor:
If I pass in a JSON URL via the param `i18n`, the month and day names do not get translated any more.
In older versions of patternslib, this is still working (e.g. older OiRA):
<img width="1609" alt="screen shot 2018-10-25 at 16 10 57" src="https://user-images.githubusercontent.com/642660/47506673-d964d180-d870-11e8-83a9-df48582f3031.png">

In the current version, this is broken (Quaive prod, new OiRA):
<img width="1609" alt="screen shot 2018-10-25 at 16 12 15" src="https://user-images.githubusercontent.com/642660/47506688-e386d000-d870-11e8-8195-8db78dcbcb5b.png">

In both cases, the JSON returned by the URL under i18n is
```
{"weekdaysShort": ["Sun", "M\u00e1n", "\u00deri", "Mi\u00f0", "Fim", "Fim", "Lau"], "months": ["Jan\u00faar", "Febr\u00faar", "Mars", "Apr\u00edl", "Ma\u00ed", "J\u00fan\u00ed", "J\u00fal\u00ed", "\u00c1g\u00fast", "September", "Okt\u00f3ber", "N\u00f3vember", "Desember"], "weekdays": ["Sunnudagur", "M\u00e1nudagur", "\u00deri\u00f0judagur", "Mi\u00f0vikudagur", "Fimmtudagur", "Fimmtudagur", "Laugardagur"]}
```

**Update** On the Patternslib site (http://patternslib.com/date-picker), the mutlilingual example is completely broken, but that seems caused by wrong URL for the JSON file. Otherwise it should be possible to reproduce the bug there (by fixing the URL).

pysailor:
Hi @fulv - in case you have time, feel free to pick this up and assign yourself to it ;)

pilz:
Hi @fulv in case you are looking into this right now. I just made a fix for pat-display-time where I switched to lazy loading of locales. If the date picker also could retrieve its translations from moment.js, we could reuse this mechanism. 
https://github.com/ploneintranet/pat-display-time/blob/master/src/pat-display-time.js#L60
Note that this requires building chunks for lazyloading as described here: https://medium.com/@michalozogan/how-to-split-moment-js-locales-to-chunks-with-webpack-de9e25caccea

pysailor:
Hi @fulv, I don't suppose you've had time to look at this issue?

fulv:
[...]

PR #630, opened 2018-11-20
Fixes issue #629

fulv:
Commit 9c98f5919c7b51cc41a5a1164773bd4af2cf8f20 introduced this bug, because `getJSON` is asynchronous.  This commit reverts a few line changes to handle this case properly.


fulv:
Those pins are only temporary since they are all going to be upgraded when the big jquery 3 merge happens.

fulv:
Please review!

fulv:
🥂

PR #637, opened 2019-02-15
Don't send a pat-update if links are marked current

pilz:


pilz:
@cornae is thinking about occurrences where this might have an effect. Tooltips that are repositioned, notifications perhaps, but we can't really pinpoint any real reason. We opt for merging this and testing the prototypes for scroll issues.
This could be the reason if you scroll on a page and something doesn't properly update itself.

PR #640, opened 2019-03-07
Fix pat notification

### LINKED ISSUE #639: pat-notification does not set properly the close-panel class
https://github.com/Patternslib/Patterns/blob/1464d7b4dbdfde4700677f2e5dfb769f1f72c346/src/pat/notification/notification.js#L98

ale-rt:
Fixes #639


pilz:
Muharharhar

PR #646, opened 2019-03-15
Optimize the argParserParse function

### LINKED ISSUE #645: argParserParse should be optimized
If inherit is true, it looks for the configuration in all the parents (`$el.parents()`) and then discards the one than do not have the parser attribute.
I believe this can be achieved in with a smarted query `$el.parents('[' + this.attribute + '}')` enhancing the performances.

ale-rt:
Fixes #645


ale-rt:
@pilz any comment about this PR?

PR #658, opened 2019-05-07
Fix easy vulnerabilities

fulv:


fulv:
I have also checked in an audit.log file.  It's probably going to get out of date, but right now I find it useful.

PR #659, opened 2019-05-08
Fix two more security vulnerabilities, remove http-server.

fulv:
I have removed `http-server` because it does not seem to be used, and because of https://github.com/jfhbrook/node-ecstatic/issues/259.  (http-server depends on ecstatic, which contains a vulnerability, the resolution of which is likely going to be messy).

The remaining vulnerabilities are in:

- node-sass, and I am confident this will be resolved shortly
- jasmine-webpack-plugin, and I need to look into this a bit more.  But the severity of these is medium or now.


pilz:
Why are the tests suddenly so stable?

fulv:
not so loud

pilz:
lol

PR #670, opened 2019-09-17
Add pat-fullscreen pattern

thet:
Add ``pat-fullscreen`` pattern to allow any element to be displayed in fullscreen-mode.


thet:
@pilz: ready for review. This PR probably should be "Squashed and merged".

PR #671, opened 2019-09-24
Add pat-fullscreen support for close

thet:
A second pattern ``pat-fullscreen-close`` which is triggered on ``close-fullscreen`` CSS class allows for closing the fullscreen with custom buttons.

### DIFF
--- a/src/pat/fullscreen/fullscreen-close.js
+++ b/src/pat/fullscreen/fullscreen-close.js
@@ -1,0 +1,20 @@
+define([
+    "pat-base",
+    "screenful"
+], function(Base, screenful) {
+    return Base.extend({
+        name: "fullscreen-close",
+        trigger: ".close-fullscreen",
+        init: function($el, opts) {
+            this.$el[0].addEventListener('click', function () {
+                // no prevent-default nor stop propagation to let
+                // the button also do other stuff.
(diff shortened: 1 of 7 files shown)


src/pat/fullscreen/fullscreen-close.js
```diff
@@ -0,0 +1,20 @@
+define([
+    "pat-base",
+    "screenful"
+], function(Base, screenful) {
+    return Base.extend({
+        name: "fullscreen-close",
+        trigger: ".close-fullscreen",
```
thet: Attention: this is against the convention of using the pattern name ``pat-fullscreen-close`` as trigger. But since this makes only sense in the context of this fullscreen pattern this might be OK.

cornae:
Why is this a second Pattern? It’s part of the same interaction pattern.

thet:
@cornae the close-fullscreen event handler is initialized before you inject any slides. So, after injection, the event handler is not applied. But the injection pattern re-scans the injected dom tree for any patterns and initializes them. Therefore the close-button needs to be in a own pattern.
I don't see another solution other then ones which would end up in a hack.
You can still use the ``.close-fullscreen`` CSS class.

thet:
travis test this

PR #672, opened 2019-09-24
[WIP] Babelize for latest ES, update, upgrade.

thet:


src/pat/auto-suggest/_auto-suggest.scss
```diff
@@ -492,7 +492,6 @@ disabled look for disabled choices in the results dropdown
     font-size: 0.9em;
     font-weight: normal;
 
-    &.undecided! {}
```
thet: @cornae node-sass (which is the new standard implementation of the SASS parser) complains about this.
I assume it's safe to remove, isn't it?

thet:
Closed in favor of: https://github.com/Patternslib/Patterns/pull/713

PR #674, opened 2019-09-30
Add webpack visualizer (backport from thet-updates)

thet:
Add webpack visualizer as a backport/cherry-pick from https://github.com/Patternslib/Patterns/pull/672 as we probably want this feature right away to understand the bundle size.
Done as pre-requisite of other bugfixes.

This gives us a stats.html. Open, select bundle.js (otherwise bundle.min.js is just summed up to bundle.js) and analyze. Should look like this:

Screenshot from 2019-09-30 13-31-37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@ bundle.js.map
 bundle.min.js
 bundle.min.js.map
 /chunks
+stats.json
+stats.html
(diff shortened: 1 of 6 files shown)


pilz:
Wonderful!

PR #678, opened 2019-10-02
When an inject fails because the source doesn't contain the intended …

pilz:
…selector, redirect to the target url instead of failing silently. This will show the login form in case an inject fails because the user is no longer authenticated.


pilz:
Actually this is not needed, we can solve this by setting the right http code in plone.

PR #681, opened 2019-10-07
Pat tooltip ng rebased

pilz:


pilz:
Seems we got ourselves another heisenbug. Locally all is green. And also one of the tests is, so I am merging this for now as this is simply an addition.

PR #696, opened 2020-01-03
Fix menu pattern

thet:
Followup of https://github.com/Patternslib/Patterns/pull/496 from 2017 to reduce open PRs as preparation for the Alpine City Sprint.

Changes from #496 plus:
- Makes menu pattern work.


thet:
The failing test is a heisenbug and can be ignored.

PR #699, opened 2020-01-29
[WIP] Trigger validation after value change/input

thet:
This is a draft and naive approach with untested pseudo code open for discussion!

Problem:
When running robot tests and filling out forms it can happen that pat-validation sees some inputs not valid although they are. In the following screenshot you can see that the required title field is filled-out but pat-validation didn't recognize that. It validates the field after it looses focus which does not happen in this robot test:

Screenshot from 2020-01-29 11-13-44

Solution:
Listen to the ``changed`` and ``input`` events with a timeout and validate while typing/checking.

Benefits:
Better validation handling, ***maybe*** also some heisenbug robot test fixes.

### DIFF
--- a/src/pat/validation/validation.js
+++ b/src/pat/validation/validation.js
@@ -58,20 +58,21 @@ define([
             this.errors = 0;
             this.options = parser.parse(this.$el, opts);
             this.$inputs = this.$el.find(':input[name]');
-            this.$el.find(":input[type=number]").on('keyup mouseup', _.debounce(function (ev) {
+
+            // Fire - debounced - when value of input elements changes
+            // See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event
+            this.$el.find("input:not([type=button]):not([type=reset]):not([type=submit]:not([type=checkbox]):not([type=radio])), select, textarea, *[contenteditable]:not([contenteditable=false])").on('input', _.debounce(function (ev) {
                 this.validateElement(ev.target);
             }.bind(this), 500));
-            this.$inputs.on('change.pat-validation', function (ev) { this.validateElement(ev.target); }.bind(this));
-            // formaction causes form validation to be skipped, so validate on
-            // submit button click instead.
-            this.$el.find('button[type=submit][formaction]').on('click.pat-validation', this.validateForm.bind(this));
-            this.$el.on('submit.pat-validation', this.validateForm.bind(this));
+            // See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event
+            this.$el.find("input[type=checkbox]), input[type=radio]").on('change', _.debounce(function (ev) {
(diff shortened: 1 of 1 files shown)


src/pat/validation/validation.js
```diff
@@ -58,9 +58,17 @@ define([
             this.errors = 0;
             this.options = parser.parse(this.$el, opts);
             this.$inputs = this.$el.find(':input[name]');
-            this.$el.find(":input[type=number]").on('keyup mouseup', _.debounce(function (ev) {
```
thet: did that ever fire? ``:input`` is not a valid selector AFAIK.

src/pat/validation/validation.js
```diff
@@ -58,20 +58,21 @@ define([
...
                 this.validateElement(ev.target);
             }.bind(this), 500));
-            this.$inputs.on('change.pat-validation', function (ev) { this.validateElement(ev.target); }.bind(this));
-            // formaction causes form validation to be skipped, so validate on
-            // submit button click instead.
-            this.$el.find('button[type=submit][formaction]').on('click.pat-validation', this.validateForm.bind(this));
-            this.$el.on('submit.pat-validation', this.validateForm.bind(this));
+
+            // Validate on form submit.
+            this.$el.on('submit', this.validateForm.bind(this));
```
thet: Just validate on form submit.
If invalid, cancel current submit.
Specifically handle AJAX submission (would be nice, if they also hook into standard submit events).
Do not validate on form close.

src/pat/validation/validation.js
```diff
@@ -58,20 +58,21 @@ define([
...
+            // Fire - debounced - when value of input elements changes
+            // See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event
+            this.$el.find("input:not([type=button]):not([type=reset]):not([type=submit]:not([type=checkbox]):not([type=radio])), select, textarea, *[contenteditable]:not([contenteditable=false])").on('input', _.debounce(function (ev) {
+                this.validateElement(ev.target);
+            }.bind(this), 500));
+            // See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event
+            this.$el.find("input[type=checkbox]), input[type=radio]").on('change', _.debounce(function (ev) {
                 this.validateElement(ev.target);
             }.bind(this), 500));
-            this.$inputs.on('change.pat-validation', function (ev) { this.validateElement(ev.target); }.bind(this));
```
thet: I don't know yet how a ``change.pat-validation`` event is triggered...?
Same with ``submit.pat-validation`` and ``click.pat-validation``...?

src/pat/validation/validation.js
thet: found the answer: https://api.jquery.com/input-selector/

thet:
Fixed with: https://github.com/Patternslib/Patterns/pull/936

PR #704, opened 2020-03-12
[WIP] Updating navigation pattern

thet:
Navigation:
- Do not set the ``.navigation-in-path`` class for elenents already marked with ``.current``.
- Allow configuration of "in path" and "current" classes via ``in-path-class`` and ``current-class`` configuration options.
- Code cleanup.


thet:
@pilz that was quick :)

pilz:
Because it broke because of me. That always triggers me :)

PR #706, opened 2020-03-17
more pat-navigation fixes

thet:
Loads of bugfixes plus added tests+docs plus:

  - Allow configuration of ``item-wrapper`` element on which ``in-path-class`` or ``current-class`` are set. Defaults to ``li``.
  - Set current class also on parent item-wrapper if not set.
  - Fix case where current class was not set when not present on load.
  - Set current class also when DOM elements are inserted or removed.


pilz:
Test Failures were unrelated.

PR #709, opened 2020-03-27
Modal: allow programmatic initialization

thet:
Modal: Allow passing of parameters ``url`` and ``trigger`` to pat-inject. This allows programmatic instantiation of a modal with injected content.

I have a usecase where I need to open a modal after clicking on a link of an external JavaScript app (ONLYOFFICE in that case). Therefore, I cannot use the traditional way of invoking the modal by attaching the ``pat-modal`` class on a link including options but I need to programmatically invoke it in an event handler. For that to work, I need to pass the ``url`` and ``trigger`` options to pat-inject. This is an example of how I do it:

```
    var modal = patterns.patterns.modal.init(
      $('<span></span>'),
      {
        url: 'http://0.0.0.0:3001/src/pat/modal/index.html#modal_content',
        trigger: 'autoload'
      }
    );
```
The dummy ``<span>`` is needed - otherwise core patternslib throws an error.

### DIFF
--- a/src/pat/modal/modal.js
+++ b/src/pat/modal/modal.js
@@ -11,6 +11,8 @@ define([
     parser.addArgument("closing", ["close-button"], ["close-button", "outside"], true);
     parser.addArgument("close-text", 'Close');
     parser.addArgument("panel-header-content", ":first:not(.header)");
+    parser.addArgument("url", null);  // inject-url
+    parser.addArgument("trigger", null);  // inject-trigger
 
     return Base.extend({
         name: "modal",
@@ -35,6 +37,13 @@ define([
                 target: "#pat-modal",
                 "class": "pat-modal" + (this.options["class"] ? " " + this.options["class"] : "")
             };
+            if (this.options.url) {
+                opts.url = this.options.url;
+            }
+            if (this.options.trigger) {
+                opts.trigger = this.options.trigger;
+            }
+
             // if $el is already inside a modal, do not detach #pat-modal,
             // because this would unnecessarily close the modal itself
             if (!this.$el.closest("#pat-modal")) {
--- a/src/patterns.js
+++ b/src/patterns.js
@@ -99,6 +99,7 @@ define([
     // Since we are in a non-AMD env, register a few useful utilites
     var window = require("window");
     window.jQuery = $;
+    window.patterns = registry;
     require("imports-loader?this=>window!jquery.browser");
 
     $(function () {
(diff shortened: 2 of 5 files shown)


pysailor:
Hi @thet ,
in order to give Cornelis some background on this, can you please describe the scenario that you are addressing, and why you need this change?

src/pat/modal/modal.js
```diff
@@ -11,6 +11,8 @@ define([
     parser.addArgument("closing", ["close-button"], ["close-button", "outside"], true);
     parser.addArgument("close-text", 'Close');
     parser.addArgument("panel-header-content", ":first:not(.header)");
+    parser.addArgument("url", null);  // inject-url
+    parser.addArgument("trigger", null);  // inject-trigger
```
thet: I was thinking of naming these arguments ``inject-url`` and ``inject-trigger`` but found that this is confusing. Other opinions?

src/patterns.js
```diff
@@ -99,6 +99,7 @@ define([
     // Since we are in a non-AMD env, register a few useful utilites
     var window = require("window");
     window.jQuery = $;
+    window.patterns = registry;
```
thet: This is only to easily instantiate the modal pattern via ``patterns.patterns.modal``. We have the same in https://github.com/quaive/ploneintranet.prototype/blob/01e4cb5c3ac2690816d5e1716ffa32c5dafef7bf/bundle-config.js#L88

cornae:
Hi @thet, I can partially follow what you are explaining but not completely. I think there are too many details to look at here in a thread, also because I missed the entire process. 

Could we do a conf call with you, I and @pilz to go through it?

pilz:
Hi Cornelis, yes, let's talk. As a first step, here is some background info on this:

In Quaive we integrate OnlyOffice so that users can edit word/excel/powerpoint documents directly. OnlyOffice comes with a nice feature where you can do mail merges. You open a word document, click the mail merge button and pick an excel file to connect as data source. That excel file comes from quaive and onlyoffice requires us to open a view on quaive to pick the excel file. 

As a proof of concept, we are opening a modal with a list of the excel files of the current
workspace so that OnlyOffice can pick its source from. 
OnlyOffice gives us only a hook in JavaScript where we can open the modal and no classical classical page load is done which would initialize the pat-modal.

So by adding the parameters to initialise the modal from javascript, we can click a button in OnlyOffice and a patterns modal shows to list the excel files.

cornae:
Thanks. That gives already some context. When do you want to plan the call?

thet:
Superseded by: https://github.com/Patternslib/Patterns/pull/709

PR #716, opened 2020-05-07
Rework push message support

thet:
Rework push message support for the STOMP message protocoll instead of backends instead of WAMP.
We are using the RabbitMQ message broker for push support instead of crossbar.io.


src/core/push_kit.js
```diff
@@ -6,103 +6,110 @@
...
+], function($, stompjs) {
+
+const push_kit = {
+
+    init() {
+        document.querySelector('meta[name="patterns-push-server"]').content
+        const push_url = $("meta[name=patterns-push-server]").attr("content");
+        const push_path = $("meta[name=patterns-push-path]").attr("content");
+        const push_user = $("meta[name=patterns-push-user-TODO]").attr("content");
+        const push_pass = $("meta[name=patterns-push-pass-TODO]").attr("content");
```
pilz, line 19: We will not store a password in a key saying that this is a password, right? It is still a Todo, so we can merge, but let's discuss authentication again.

pilz:
Good for testing. Here, speed is more important to get stuff out to test beds. Merging

src/core/push_kit.js
pysailor: Exactly, @thet and I agreed earlier in a call that it is better for testing if we can pass the credentials in markup, rather than hard-coding them in the pattern. But we are aware that we don't want to put credentials in markup. We simply postponed this problem to a later stage when we have gathered experience.

thet: sure, therefore the -TODO :)
The authentication question needs to be resolved... Maybe we can use a web token (JWT) or some similar mechanism

PR #718, opened 2020-05-13
Fix initial load for transition: css

lucaswiersma:
The class visible or hidden was never set in the on init method. 
To fix I use the hideOrShow utils method in the on init instead of calling the show/hide methods. 
Subsequently we call the updateModal method which was otherwise called in the show/hide methods. these two methods will now no longer be used and therefore removed.

### DIFF
--- a/src/pat/depends/depends.js
+++ b/src/pat/depends/depends.js
@@ -45,14 +45,11 @@ define([
                 log.error("Invalid condition: " + e.message, slave);
                 return;
             }
-
             state=handler.evaluate();
             switch (options.action) {
                 case "show":
-                    if (state)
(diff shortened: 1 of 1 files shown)


pilz:
Thank you for the pull request! We'll look at it on friday.

pilz:
I can see that most of the black magic setting classes and styles are not done in the pat depends pattern. So to solve the additional issues mentioned by Cornelis, we need to investigate deeper into jQuery. For now I don't see any harm with this and will merge so that Cornelis can continue testing, but we will need to expand that pattern to address the issues mentioned in #719. NB that there is no support in depends.js for transition:css.

Issue #721, opened 2020-05-14
Pat scroll selector: top


cornae:
An important use case for pat-scroll is to use it to auto scroll to the top of a content when you come from an overview page at a larger position. Generally the designer wants to scroll to the top most position. In some designs it's hard to find a suitable element to scroll to. A proposed solution is to introduce a special value 'top': `selector: top`. The value 'top' could be used instead of a CSS selector in case no usable element is available in the design.

I've fitted the test page with an example.

thet:
Notes:
- selector: top should scroll to position 0/0
- selector should take precedence over href
- Branch with demo page: https://github.com/Patternslib/Patterns/compare/scroll-selector-top
- Prototype ticket: https://github.com/quaive/ploneintranet.prototype/issues/986

Issue #703, opened 2020-02-27
Global pat-inject 


cornae:
## Context

In the beginning when the web was void and empty we used a pat-inject statement here and there to smoothen things up. Today I fill my days writing `class="pat-inject" data-pat-inject="history: record"` on every possible link. History: record is something that we want almost every time. 

## Problem

We could make it a default and have history: no-record for the exceptions. But the pat-inject class is still needed on every link and form which is a lot of work. Also we have no control over links in rich texts written by editors. And those should use inject too for smooth operation of applications such as Quaive. 

## Proposal

I propose to introduce a 'selector' property for pat-inject. The selector property allow for pointing to the element(s) on which the inject statement should take effect. The default value is 'self' like it is now. The selector would within the context of the container element the statement is placed on. 

Example:

```html
<body class="pat-inject" 
      data-pat-inject="selector: a,form; source: #main; target: #main; history: record">
```

The above example would apply basic injection behaviour for everthing that loads content on the site. 

It should be possible to overrule the inject statement. For instance when an injection is sometimes not desired or when another combination of source and target is desired. 

To a designer it would feel natural to overrule property by property. But I think that would be too complex. I would propose to overrule pat-inject on a link or form element altogether simply by putting a class pat-inject on the element. It will accept only its own properties. 

When no injection is required on an element that matches the pat-inject selector of its container, I propose the following: `class="pat-inject" data-pat-inject="none"`.

cornae:
Note: the current pat-inject documentation mentions that a property 'selector' already exists, but I think that that's a typo and that that should be 'source'.

thet:
I like the global injection idea.
The ``selector`` documentation is a typo (fixed in PR #724) but there is really also a ``selector`` property which can be defined as default if no source and target are defined (renamed in PR #724).

PR #722, opened 2020-05-19
pat-scroll: Implement `selector:top`

thet:
Ref: https://github.com/Patternslib/Patterns/issues/721

Note:
IIRC, The original issue appeared when injecting content into a site where we want to scroll to the top of the page and not keep the old scroll position.
This problem is actually widely known - e.g. vue-router has a special `scrollBehavior` option for route-changes for this: https://router.vuejs.org/guide/advanced/scroll-behavior.html
An alternative solution would be to implement the scrolling behavior in pat-inject after the injection was done.
@cornae @pilz

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 Features
 ~~~~~~~~
 
+- pat-scroll: Implement new special `selector:top` attribute value to scroll the scroll container just to the top of the page. Ref: #721.
 - pat-scroll: To define the scrollable target search also for `overflow-x` and `overflow-y` declarations.
 - Rework push message support for the STOMP message protocoll instead of backends instead of WAMP.
   We are using the RabbitMQ message broker for push support instead of crossbar.io.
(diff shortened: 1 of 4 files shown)


pilz:
@thet we have some scroll impl in inject already, did you know https://github.com/Patternslib/Patterns/blob/master/src/pat/inject/inject.js#L501 ?

thet:
@pilz - very nice! Did not know this. Can scroll-top for pat-inject be the default? If so, then this work here might be obsolete - although it's still handy to have the option to just scroll up without any further config.

cornae:
@thet scroll: top as default for pat-inject would have unoverseable consequences. But… we could reach the same effect once we have global pat-inject. https://github.com/Patternslib/Patterns/issues/703 With global pat-inject you would only have to put pat-inject on the body tag with the scroll property set to top. 

What pat inject scroll cannot do is animate the scroll and sometimes that's more elegant. The top value is either way still very welcome for pat-scroll.

PR #724, opened 2020-05-21
pat-inject: default selector

thet:
- pat-inject: Rename undocumented ``selector`` property to ``defaultSelector``.
- pat-inject: Fix typo in docs for the ``source`` property.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 Features
 ~~~~~~~~
 
+- pat-inject: Fix typo in docs for the ``source`` property.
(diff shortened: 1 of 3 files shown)


cornae:
Hi @thet, what does default selector do exactly? 'self'? And then if you use it would it select another element as the section trigger?

The property should be named default-selector as, similar to CSS, Patterns never uses camel case for its properties.

thet:
@cornae thanks for the review! I have changed it to ``default-selector``.
If there is a ``#IDSELECTOR`` part in the url property or the ``href`` attribute, then this is used as default selector. If not, ``body`` is used. This default selector is only useed when no ``source`` or ``target`` are defined.

PR #727, opened 2020-05-27
pat-scroll-box - new pattern out of scrolling-detection

thet:
pat-scroll-box: New pattern for scrolling detection.
Replaces the previous "scroll detection" module.


cornae:
Wat is pat-scrolling-box? Didn't we we go for pat-scroll-box as the name of the pattern? If pat-scrolling-box is something that supports pat-scroll-box, shouldn't we call it something else so that we keep the pat- prefix for things that are (interaction) patterns?

thet:
that was just a typo - i corrected it.

cornae:
Ok!

PR #728, opened 2020-05-27
[WIP] failing tests for pat-scroll-box

thet:
these tests are failing and i give up. don't want to mess around with this outdated infrastructure but better fix this in es6-core.

:collision:


pilz:
I agree. Let's Cornelis be our test suite on this one.

PR #731, opened 2020-06-17
[WIP] ES6+ rework

thet:
ECMAScript 6+ rework:
- Remove RequireJS, use ES6 import feature
- Use Jest as test framework
- Split core and lib out to https://github.com/Patternslib/patternslib-core


thet:
Archived. core-split isn't easily possible yet. Jest breaks.

PR #732, opened 2020-06-22
fix rebaseURL to work only with correct base URLs

thet:
core/utils rebaseURL: Do not rebase if the base URL isn't absolute or doesn't start with an URL scheme.

Note: the method ``_rebaseHTML_via_HTMLParser`` is unused.
I will remove this method and ``lib/htmlparser`` in https://github.com/Patternslib/Patterns/pull/731

### DIFF
--- a/src/pat/inject/inject.js
+++ b/src/pat/inject/inject.js
@@ -827,17 +827,22 @@ define([
                 "$1src=\"\" data-pat-inject-rebase-$2="
             ).trim()).wrapAll("<div>").parent();
 
-            $page.find(Object.keys(inject._rebaseAttrs).join(",")).each(function() {
-                var $this = $(this),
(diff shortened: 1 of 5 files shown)


src/pat/inject/inject.js
```diff
@@ -827,17 +827,22 @@ define([
...
-                if (value && value.slice(0, 2) !== "@@" && value[0] !== "#" &&
-                    value.slice(0, 7) !== "mailto:" && value.slice(0, 11) !== "javascript:") {
-                    value = utils.rebaseURL(base, value);
-                    $this.attr(attrName, value);
-                }
-            });
+            if (
+                base.indexOf("://") !== -1 ||
+                base[0] === "/"
+            ) {
```
thet, line 833: mainly for performance reasons - if there is no correct base url there is no need to parse the HTML tree

PR #736, opened 2020-06-29
Simplify package.json, remove phantomjs

thet:
- Upgrade moment and moment-timezone
- Simplify package.json and remove unused.
- Remove PhantomJS - we're using ChromeHeadless already.
- npm install


thet:
@pilz fails due to flaky tippy test.

PR #740, opened 2020-07-23
pat-validation: Added the possibility to check for fields equality

ale-rt:
This is particularly interesting for password or email confirmation fields

### DIFF
--- a/src/pat/validation/documentation.md
+++ b/src/pat/validation/documentation.md
@@ -8,7 +8,7 @@ The validation pattern is triggered by a single class `pat-validation` on the fo
 This pattern has several advantages over standard HTML 5 form validation:
 
 - it supports older browsers
-- it uses simple documented HTML markup to allow non-browser-specific styling of error messages 
+- it uses simple documented HTML markup to allow non-browser-specific styling of error messages
 - it supports extra validation rules
 
 ### The following attributes may be used.
@@ -68,6 +68,7 @@ Error messages can also be overridden on a per-field basis, for example:
 | Property          | Description                                                                        | Default | Type |
 |-------------------|------------------------------------------------------------------------------------|---------|------|
 | disable-selector  | A selector for elements that should be disabled when there are errors in the form. |         | CSS Selector |
+| equality          | Field-specific. The name of another input this input should equal to (useful for password confirmation). | | String |
 | message-date      | The error message for date fields.        | This value must be a valid date | String |
 | message-datetime  | The error message for datetime fields.    | This value must be a valid date and time | String |
 | message-email     | The error message for email fields.       | This value must be a valid email address | String |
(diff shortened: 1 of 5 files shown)


cornae:
Very useful!

src/pat/validation/documentation.md
```diff
@@ -68,6 +68,7 @@ Error messages can also be overridden on a per-field basis, for example:
 | Property          | Description                                                                        | Default | Type |
 |-------------------|------------------------------------------------------------------------------------|---------|------|
 | disable-selector  | A selector for elements that should be disabled when there are errors in the form. |         | CSS Selector |
+| equality          | Field-specific. The name of another input this input should equal to (useful for password confirmation). | | String |
```
thet, line 71: Please also add an entry for ``message-equality``.

src/pat/validation/validation.js
```diff
@@ -163,6 +166,18 @@ define([
                     constraint.numericality.onlyInteger = true;
                 }
             }
+
+            // Handle fields equality
+            if (opts.equality && input.value) {
+                this.$el.find("[name=" + opts.equality + "]").each(
+                    function (idx, el) {
+                        if (input.value !== el.value) {
+                            constraint.equality = {'attribute': opts.equality, 'message': '^'+opts.message["equality"]};
```
thet, line 175: In my understanding of http://validatejs.org/#validators-equality the ``attribute`` attribute shouldn't be neccessary.

thet: Do you know what the ``'^`` is for? it looks strange, but I see all the other ``message-*`` attributes have it too.

src/pat/validation/validation.js
```diff
@@ -163,6 +166,18 @@ define([
                     constraint.numericality.onlyInteger = true;
                 }
             }
+
+            // Handle fields equality
+            if (opts.equality && input.value) {
```
thet, line 171: I think the ``&& input.value`` check should be removed. You probably still want to check for equality if no value was given. Maybe instead of the ``&& input.value`` we should enforce presence via the ``required`` attribute?

src/pat/validation/validation.js
ale-rt: > In my understanding of http://validatejs.org/#validators-equality the `attribute` attribute shouldn't be neccessary.

If I remove it the tests fail. I think `attribute` is mandatory when you also specify a `message`.

> Do you know what the '^ is for? it looks strange, but I see all the other message-* attributes have it too.

No idea, I put it there to be consistent with the already existent code :)

src/pat/validation/documentation.md
ale-rt: Added!

src/pat/validation/validation.js
ale-rt: I removed the check

ale-rt:
I implemented all the requested changes except https://github.com/Patternslib/Patterns/pull/740#discussion_r460289911 because either I did not understand how to do it or it is not possible to omit the `attribute` once a message is specified.
So this one is ready to be rereviewed.

PR #742, opened 2020-08-10
pat-validation: Allow for HTML5 style required attributes without a value

thet:
The HTML5 specs allow for an ``required`` attribute without a value. We're not supporting this in pat-validation until now. This PR fixes this.

Now we can use:
```
<input type="text" name="req1" required="required" />
```
as well as 
```
<input type="text" name="req1" required />
```

See:
https://html.spec.whatwg.org/multipage/input.html#attr-input-required
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,11 +4,11 @@
 
 Features
 ~~~~~~~~
+- pat-validation: Allow for HTML5 style ``required`` attributes without a value.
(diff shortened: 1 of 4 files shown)


pysailor:
> But I don't know how to trigger re-running the tests via a GH comment.
Of course, directly in travis, this is possible. I just restarted the build

PR #743, opened 2020-08-13
[WIP] pat-gallery: Add feature to auto-wrap <img> tags within pat-gallery.

thet:
From the docs:

Sometimes it's not possible to modify content with images to fullfill pat-gallery's markup requirements.
In this case you can auto-add these images and also change the ``src`` value via search/replace.

```
    <nav class="pat-gallery" data-pat-gallery="item-selector: a.add-to-gallery; autoadd-enabled: true; autoadd-src-replace-from: thumb-; autoadd-src-replace-to: full-;">
      <a href="full-1.jpg" class="add-to-gallery"><img src="thumb-1.jpg"/></a>
      <img src="thumb-2.jpg"/>
      <img src="thumb-3.jpg"/>
    </nav>
```

This will result in:
```
    <nav class="pat-gallery" data-pat-gallery="item-selector: a.add-to-gallery; autoadd-enabled: true; autoadd-src-replace-from: thumb-; autoadd-src-replace-to: full-;">
      <a href="full-1.jpg" class="add-to-gallery"><img src="thumb-1.jpg"/></a>
      <a href="full-2.jpg" class="autoadd-gallery-selector"><img src="thumb-2.jpg"/></a>
      <a href="full-3.jpg" class="autoadd-gallery-selector"><img src="thumb-3.jpg"/></a>
    </nav>
```

Please note the usage of the special class ``autoadd-gallery-selector`` to mark images to be added to the gallery. This is because we cannot determine the class from the pat-gallery configuration (the selector can also include id and element selectors).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 Features
 ~~~~~~~~
+- pat-gallery: Add feature to auto-wrap <img> tags within pat-gallery.
 - pat-validation: Allow for HTML5 style ``required`` attributes without a value.
 - pat-validation: Added the possibility to check for fields equality
 - pat-validation: Dont use :input jquery extension for better performance
--- a/src/pat/gallery/documentation.md
+++ b/src/pat/gallery/documentation.md
@@ -47,6 +47,15 @@ If you just want individual images to be opened in the gallery overlay but want 
       <a class="pat-gallery" href="images/full/2.jpg"><img src="images/thumb/2.jpg"/></a>
     </nav>
 
+Sometimes it's not possible to modify content with images to fullfill pat-gallery's markup requirements.
+In this case you can auto-add these images and also change the ``src`` value via search/replace.
+
+    <nav class="pat-gallery" data-pat-gallery="item-selector: a.add-to-gallery; autoadd-enabled: true; autoadd-src-replace-from: thumb-; autoadd-src-replace-to: full-;">
+      <a href="full-1.jpg" class="add-to-gallery"><img src="thumb-1.jpg"/></a>
+      <img src="thumb-2.jpg"/>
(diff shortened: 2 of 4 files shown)


cornae:
Hi @thet Can you demo this in the design call tomorrow so that I understand it? I don't fully understand it, but I already see things that we don't/shouldn't do with Patterns. Including the use of booleans in properties. Please try to start the plans for changes to a pattern always in the Design call and using a design process. 

NB At this moment fixes to pat-calendar are for more important.

thet:
Hi, this refers to https://github.com/quaive/ploneintranet.prototype/issues/993

There is already a boolean option in pat-gallery: ``loop``.

The reason for this feature is, that we want pat-gallery be active also for images added via the "redactor" html editor where we cannot easily change the markup produced by it.

This is a quote from the development ticket:
```
In a discussion between Ale, Thet and Alex, we found that we would like to enhance pat-gallery so that it can scan for all images in its content and enhance them to become the gallery. That way we don't have to rewrite the markup produced by redactor. 
```
I can show you tomorrow the effects of this.

thet:
By the way, the pat-calendar issues are right next on my todo list.

thet:
Reopening with "hold-merge" label to not loose track of this feature.
We're going to implement it by altering the HTML output server-side ("plone.transformchain"), but maybe we want this thing baked ready later. At least it also gives an idea on what to do with the plone.transformchain method.

cornae:
If it turns out that server side will take more hours than anticipated, we should revert to this plan, but then we need to give more thought to the syntax of the properties, guessing of filenames for thumbs and all other aspects of this change, with all the disciplines around the table.

thet:
Closing in favor of: https://github.com/Patternslib/Patterns/pull/750

PR #749, opened 2020-08-25
Include hidden input.pat-autosuggest in validation

### LINKED ISSUE #683: pat-validation ignores "required" on pat-autosuggest
When adding an input.pat-autosuggest with a "required" attribute to a form.pat-validation element, it is ignored and the form can be submitted without entering/selecting a value.

Minimal example:

```
        <form class="pat-validation">
            <input type="text" required="required" name="input1" />
            <input type="text" class="pat-autosuggest" required="required" name="input2" />
            <button type="submit">Submit</button>
        </form>
```

Submitting the form does not work when omitting the first input, but it does work when filling in the first but omitting the second (autosuggest) input.

reinhardt:
Fixes #683

### DIFF
--- a/src/pat/validation/tests.js
+++ b/src/pat/validation/tests.js
@@ -23,7 +23,27 @@ define(["pat-registry", "pat-validation"], function(registry, pattern) {
                 '<form class="pat-validation">'+
                 '<input type="text" name="name">'+
                 '</form>');
+            pattern.init($el);
             $el.find(':input').trigger('change');
+            expect($el.find('em.warning').length).toBe(0);
+        });
+
+        it("validates required pat-autosuggest inputs on form submit", function() {
+            var $el = $(
+                '<form class="pat-validation">'+
+                '<input type="hidden" name="name" required="required" class="pat-autosuggest" data-pat-autosuggest="words: one, two">'+
+                '</form>');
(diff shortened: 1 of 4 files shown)


src/pat/validation/tests.js
```diff
@@ -27,6 +27,24 @@ define(["pat-registry", "pat-validation"], function(registry, pattern) {
...
+            pattern.init($el);
+            $el.submit();
+            expect($el.find('em.warning').length).toBe(1);
+
+            // Check that validation passes when required is not here.
+            $el = $(
+                '<form class="pat-validation">'+
+                '<input type="hidden" name="name" class="pat-autosuggest" data-pat-autosuggest="words: one, two">'+
+                '</form>');
+            $el.submit();
```
reinhardt, line 44: Do we need another `pattern.init($el)` here?

thet: good catch, yes it does. checked it and added it myself.

reinhardt: If yes then we probably also need it in the previous test which I copied this from...

thet: perfect, also added it there.

thet (approved):
LGTM

PR #750, opened 2020-08-25
pat-gallery: Allow images to directly be added to the gallery.

thet:
Follow up of: https://github.com/Patternslib/Patterns/pull/743

Clicking on directly added images also open them in the gallery UNLESS they are wrapped by an anchor element itself. Demoed in index.html.

No tests, because there isn't a test suite for the gallery yet.


pilz:
awaiting alignment with specs

PR #766, opened 2020-10-22
Thet modernize inject

thet:
... from an older PR


thet:
Obsolete due to recent activity and: https://github.com/Patternslib/Patterns/pull/788

PR #774, opened 2020-11-05
Tooltip fixes

thet:
Ref: https://jira.syslab.com/browse/SCR-660

1)  Introduce new option ``arrow-padding`` to define the padding of the box arrow from the corners of the tooltip.

2) Set content when mounting to avoid positioning problems.

@cornae the second point should fix the problem with the misplaced tooltip when opening the personal menu.

Regarding mentions tooltips in the activities stream comments box, which reaches far below the fold: I want to re-mention the already discussed fix, because I think this one would make the situation better than it is without me adding any functionality.
If we set the following in CSS (or something similar):
```
.tippy-box .tippy-content {
    max-height: calc(50vh - 50px);
    min-height: 20em;
}
```
We get this result:
Screenshot from 2020-11-05 21-42-40

You can even restrict it to specific tooltips if, for example, the tags should open bigger: ```.tags .tippy-box .tippy-content { ...```

Of restrict it for specific screen sizes with media queries.

There is still the option of calculating the max-height based on the scroll position but I doubt that this would give us much better results.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,6 +37,7 @@
 -   pat calendar: Store view, date and active categories per URL, allowing to individually customize the calendar per page.
 -   pat tooltip: Use tippy v6 based implementation.
 -   pat tooltip: Introduce new option ``arrowPadding`` to define the padding of the box arrow from the corners of the tooltip.
+-   pat tooltip: set content when mounting to avoid positioning problems.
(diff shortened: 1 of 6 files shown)


thet:
@cornae another solution which I just remembered: You can define multiple positions. In that case it would be something like:
```data-pat-tooltip="position: tl, lm; ...```

Leads to:
Screenshot from 2020-11-05 22-09-32

Issue #752, opened 2020-08-27
Pat-tabs open more tabs on click


cornae:
Currently the more tabs span of pat-tabs responds to hover and on most touch devices it responds to tap (CSS only). Sometimes a click would be more appropriate on desktops too. This could be achieved by alternating between the classes `open` and `closed` on the pat-tabs element when the user clicks on the span. The rest can then be done with CSS.

thet:
Fixed with: https://github.com/Patternslib/Patterns/pull/773

cornae:
Hi @thet, thanks for this. I noticed that the class went to the span element instead of to the pat-tabs element. I can work with that too, albeit that having the Classon the pat-tabs element would have me given more freedom. Was it not possible according to spec?

cornae:
Second question, I notice that the class for open is 'opened' instead of 'open'. Was there a special reason for that?

Issue #751, opened 2020-08-27
Pat-tabs available width calculation issues


cornae:
There are currently issues with how pat-tabs calculates the available width for itself. It seems to depend on the moment when it calculates the available width. This can both result in taking up less width that is available, or sometimes claiming more space than there is.

### Example 1 

https://quaive.cornelis.amsterdam/library.html 

Resize the browser window back and forth with a few pixels and watch the tabs grow and shrink and grow again.

### Example 2

https://quaive.cornelis.amsterdam/workspaces/open-market-committee/index

Click on the cog wheel. Sometimes all four tabs will shown, sometimes only 2 or 3, even if there is space for 4. Also it happens that all 4 tabs are shown and the more tabs span is still injected in the markup even if the span is empty. It's difficult to reproduce this. It all seems to depend on race condition. Speed of the line, speed of the browser, etc.

## Ideas

The measuring of the available width would probably need to happen in a certain order:

1. Hide the element that has pat-tabs on it. 
2. Measure the available space. That is the content width (so the inner width) of the parent element of the element that has the class .pat-tabs.
3. Who pat-tabs with a max-width inline style property on it that has the parent's content width as its value.
4. Put a class 'tabs-ready' on the pat-tabs element once everything is done.
5. Repeat when mutations are observed in the DOM.

thet:
Fixed with: https://github.com/Patternslib/Patterns/pull/773

PR #775, opened 2020-11-05
Scr 661 fixes

thet:
1) Fix classname to "open" and attach it to the ``pat-tabs`` element instead of ``extra-tabs``.

2) Fix to debounce initial tabs adjustment and place it after initialization of the ResizeObserver. This fixes wrong size calculation when pattern is initialized but layout not settled.

Refs:
https://jira.syslab.com/browse/SCR-661
https://github.com/Patternslib/Patterns/issues/752
https://github.com/Patternslib/Patterns/issues/751
https://github.com/quaive/ploneintranet.prototype/issues/1111
https://github.com/quaive/ploneintranet.prototype/issues/1112


cornae:
Hi @thet, I see now the right class in the right place. Works great. I already adapted the CSS. But clicking on anything else doesn't set the class open back to closed yet.

Issue #761, opened 2020-10-07
pat-autofocus in an iframe context


cornae:
When pat-autofocus is used on a page that runs in an iframe/object context, then most likely it's not desired to apply autofocus as it takes control away from the host page. Ideally, pat-autofocus would detect that it's running in an an iframe/object context and in that case not execute.

pilz:
The reason for this is that if we have a help page with many images, the page starts hopping wildly around, moving fields into view as they appear on loading. Seen on the oira help pages.

PR #778, opened 2020-11-11
Do not autofocus in iframes + other fixes.

thet:
- Do not autofocus in iframes. Fixes: #761.

Plus:
- Implement documented behavior to not focus on prefilled element, if there is another autofocus element which is empty.
- Instead of calling autofocus for each element call it only once.
- Register event handler only once.

### DIFF
--- a/src/pat/autofocus/autofocus.js
+++ b/src/pat/autofocus/autofocus.js
@@ -6,6 +6,11 @@ export default Base.extend({
     trigger: ":input.pat-autofocus,:input[autofocus]",
 
     init() {
+        if (window.self !== window.top) {
+            // Do not autofocus in iframes.
+            return;
+        }
+
         this.setFocus(this.trigger);
         $(document).on("patterns-injected pat-update", (e) => {
             this.setFocus($(e.target).find(this.trigger));
--- a/src/pat/autofocus/autofocus.js
+++ b/src/pat/autofocus/autofocus.js
@@ -1,34 +1,21 @@
-/**
- * Patterns autofocus - enhanced autofocus form elements
- *
- * Copyright 2012-2013 Simplon B.V. - Wichert Akkerman
- */
 import $ from "jquery";
-import registry from "../../core/registry";
+import Base from "../../core/base";
 
-var autofocus = {
+export default Base.extend({
     name: "autofocus",
     trigger: ":input.pat-autofocus,:input[autofocus]",
 
-    init: function init() {
+    init() {
         this.setFocus(this.trigger);
-        $(document).on("patterns-injected", function (e) {
-            autofocus.setFocus($(e.target).find(autofocus.trigger));
-        });
-        $(document).on("pat-update", function (e) {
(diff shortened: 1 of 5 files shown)


src/pat/autofocus/autofocus.js
```diff
@@ -1,34 +1,29 @@
...
-        }, 10);
-    },
-};
 
-registry.register(autofocus);
-export default autofocus;
+    setFocus(target) {
+        const $all = $(target);
+        const visible = [...$all].filter((it) => $(it).is(":visible"));
+        const empty = visible.filter((it) => it.value === "");
```
thet, line 23: @pilz: this fix is for the documented behavior (in documentation.md + the index.html) to not autofocus if the input is not empty and there is another autofocus element.

src/pat/autofocus/autofocus.js
```diff
@@ -1,34 +1,36 @@
...
-};
 
-registry.register(autofocus);
-export default autofocus;
+    setFocus(target) {
+        // Clear previously scheduled tasks.
+        // setFocus normally operates on the whole DOM anyways and if scoped
+        // (via the event handler in init) the last one wins.
+        // This is OK since only one element can have the focus.
+        clearTimeout(scheduled_task);
```
thet, line 27: @pilz this is an optimization to call autofocus only once and not multiple times for each pat-autofocus element which can be found in the DOM.

src/pat/autofocus/autofocus.js
```diff
@@ -1,34 +1,44 @@
...
-    setFocus: function (target) {
-        var $all = $(target);
-        var $visible = $all.filter(function () {
-            if ($(this).is(":visible")) return true;
-        });
-        setTimeout(function () {
-            $visible.get(0) && $visible.get(0).focus();
-        }, 10);
+
+        if (!registered_event_handler) {
```
thet, line 19: @pilz: We want the global event handler registered only once.

src/pat/autofocus/autofocus.js
```diff
@@ -1,34 +1,44 @@
...
+            registered_event_handler = true;
+        }
     },
-};
 
-registry.register(autofocus);
-export default autofocus;
+    setFocus(target) {
+        // Exit if task is scheduled. setFocus operates on whole DOM anyways.
+        if (scheduled_task) {
```
thet, line 30: @pilz: If there is already a setFocus task scheduled, exit. No need to do it multiple times as this method operates on the whole DOM tree and is not scoped for the element itself.

src/pat/autofocus/autofocus.js
```diff
@@ -1,34 +1,44 @@
...
+let scheduled_task = null;
+let registered_event_handler = false;
+
+export default Base.extend({
     name: "autofocus",
     trigger: ":input.pat-autofocus,:input[autofocus]",
 
-    init: function init() {
+    init() {
+        if (window.self !== window.top) {
```
thet, line 12: @pilz: requested fix.

PR #780, opened 2020-11-16
pat calendar: Do not modify original event sources config.

thet:
Fixes problem with doubling up events when visiting different fullcalendar pages with different urls via pat-inject.

### DIFF
--- a/src/pat/calendar/calendar.js
+++ b/src/pat/calendar/calendar.js
@@ -57,8 +57,6 @@ parser.addArgument("event-color", "blue");
 
 parser.addArgument("url", null);
 parser.addArgument("event-sources", [], undefined, true);
-parser.addArgument("event-sources-classes", [], undefined, true);
-parser.addArgument("event-sources-active", [], undefined, true);
 //parser.addArgument("add-url", null);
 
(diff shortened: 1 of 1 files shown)


src/pat/calendar/calendar.js
```diff
@@ -175,21 +172,15 @@ export default Base.extend({
             config.timeZone = timezone;
         }
 
-        const sources = opts.event.sources || [];
+        const sources = [...(opts.event.sources || [])];
```
thet, line 175: this is the actual fix - this copies the values instead of referencing them.

src/pat/calendar/calendar.js
```diff
@@ -57,8 +57,6 @@ parser.addArgument("event-color", "blue");
 
 parser.addArgument("url", null);
 parser.addArgument("event-sources", [], undefined, true);
-parser.addArgument("event-sources-classes", [], undefined, true);
-parser.addArgument("event-sources-active", [], undefined, true);
```
thet, line 61: I removed this as this is a too complicated way for an early idea on how to configure multiple calendars. Removed that and simplified the code below a bit.

PR #784, opened 2020-12-03
pat-inject: Fix documentation about special target self and demo that behavior

thet:
@cornae I added a demo test case to illustrate the behavior you reported to me.
I can't reproduce it - it seems to work just fine.
Maybe the markup needs to be adapted to illustrate the misbehavior?

I need some more info to create a ticket for the problem.


cornae:
I'm trying to test but make doesn't work anymore.

image

@thet Can you help out?

thet:
@cornae I have changed the Makefile in a way, that it automatically installs "yarn" if it's not available. Please pull and try again.

thet:
If it works I will update all other repos where yarn is used also...

cornae:
Thanks, I'm going to try.

PR #790, opened 2020-12-22
Rebase URLs in pattern options.

thet:
- core parser, core mockup-parser, core base, core registry: modernize code
- code dom: Make find_parents more robust against DocumentFragments with no parent.
- core utils: Add jqToNode to return a DOM node if a jQuery node was passed
- Fix tests: Add missing import for jquery-ext where it's needed.
- pat autofocus: Avoid jQuery extended CSS selector syntax.
- pat slides: Avoid jQuery extended CSS selector syntax.
- __pat inject: Rebase URLs in pattern options.__


thet:
This PR got quite massive!
It contains a number of micro-commits (i have the rule of one extra commit per pattern) PLUS it contains a lot of JS modernization in src/core (done on my trainride to vienna).

PR #792, opened 2021-01-08
[WIP] Mockup updates

thet:
From the Plone 6 UI Sprint.

/cc @MrTango


thet:
That PR or individual comits of it are already merged since 4.0.0.

PR #806, opened 2021-02-08
pat-calendar: pat-inject is only set event has a url.

thet:
I noticed a pat-inject loading icon on events without a URL parameter, which then are rendered as `<a>` elements without a `href` - see screenshot.
This PR fixes that - it only sets pat-inject if the event has a url configured.
pat-switch is still active on these elements.

@cornae do you agree with this changed behavior?

Use cases might be where users do not have the rights to view the detail view of an event (like in our absences app).

Screenshot from 2021-02-08 19-31-45


cornae:
Yes very good. I didn’t think about this usecase. Make sure to document this probably also in the code itself so that we still know why we did this in a few years from now :)

PR #810, opened 2021-02-17
pat calendar: Support pat-tooltip

thet:
pat calendar: Support `pat-tooltip` on events with a URL via `pat-tooltip-source` set to `ajax`.

@cornae I gave this feature a second shot, as we need it for the absences app.
I followed your suggestion in https://github.com/quaive/ploneintranet.prototype/pull/1071#issuecomment-683847959 but implemented currently only `pat-tooltip-source` which can be set to "ajax" as only possible value (content or anything else doesn't make sense I guess...).

I did not implement `pat-tooltip-position-list` or other options yet - I'll see how good the default positioning helps us with our requirement.

@cornae I you need any other options, let me know

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,8 @@
 -   pat calendar: Support `pat-switch` for rendered events via some configuration options.
 -   pat calendar: Support `pat-tooltip` on events with a URL via `pat-tooltip-source` set to `ajax`.
 -   pat calendar: Store view, date and active categories per URL, allowing to individually customize the calendar per page.
+-   pat calendar: Support `url` in the event JSON model additionally to `@id`.
+    The unique identifier is often not semantically correct for a URL to the item, especially when we want to call a specific view.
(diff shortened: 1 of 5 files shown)


cornae:
I think we should most effectively discuss this in a call. I have a small stack of items in need of discussion including.

1. tooltip 
2. target: self, something that came up that I need to understand
3. The list of items that is kept on https://github.com/Patternslib/Patterns/projects/1 <https://github.com/Patternslib/Patterns/projects/1>

@pilz Are you ready to initiate the call that we talked about for #3? We can then do #1 and #2 at the same time. If you and Johannes are to busy for #3 currently, then I can also setup a separate call with Johannes on #2 and #3.

thet:
@cornae can we schedule a meeting tomorrow morning? I'm available from 08:30 on.

thet:
Meeting with cornelis, the changes are OK. merging.

PR #814, opened 2021-03-08
Revert "Do not apply filter, if filter is empty.", as this leads to unexpected UX.

thet:
Commit was made due to no events shown in a calendar without filters
applied.
The real fix is to not include any filtering controls.
The code already supports showing all events when no filtering controls
are available.

Revert "pat-calendar: Do not apply filter, if filter is empty."

This reverts commit 2acb2c7cf8230b58917363412fa6b81cd05e1eed.

### DIFF
--- a/src/pat/calendar/calendar.js
+++ b/src/pat/calendar/calendar.js
@@ -95,7 +95,7 @@ export default Base.extend({
         agendaDay: "timeGridDay",
     },
     dayNames: ["su", "mo", "tu", "we", "th", "fr", "sa"],
-    active_categories: [],
+    active_categories: null,
     parser: parser,
 
     async init($el, opts) {
@@ -386,7 +386,7 @@ export default Base.extend({
 
     filter_event(event) {
         let show = true;
-        if (this.active_categories.length) {
+        if (this.active_categories !== null) {
             // intersection
             show =
                 this.active_categories.filter((it) =>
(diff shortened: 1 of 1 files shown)


cornae:
So does that mean that te problem is maybe in the HTML and not in the .js? By default, all calendars should be ticked in the HTML.

thet:
The absences app do not have any filters, but in the html layout a hidden checkbox is present. That did not match with any of the absences, so no absences were show in the calendar.

The problem was:
- The commit which I'm reverting here - had the effect that if no checkboxes are checked, the filter is not applied and no events are shown.
- There is a change necessary in ploneintranet.prototype as well as quaive.app.absences. It removes the hidden filter checkbox, so that the pattern doesn't apply any filters. The hidden filter would show all events with the css class "cal-cat-all", which our events do not have yet.

I'll finish the PRs and will ping you again!

cornae:
> Op 8 mrt. 2021, om 11:43 heeft Johannes Raggam <notifications@github.com> het volgende geschreven:
> 
> 
> The absences app do not have any filters, but in the html layout a hidden checkbox is present. That did not match with any of the absences, so no absences were show in the calendar.
> 
> The problem was:
> 
> The commit which I'm reverting here - had the effect that if no checkboxes are checked, the filter is not applied and no events are shown.
It’s correct that if no checkbox are checked, that no events are shown. That’s the intended behaviour. 
> There is a change necessary in ploneintranet.prototype as well as quaive.app.absences. It removes the hidden filter checkbox, so that the pattern doesn't apply any filters. The hidden filter would show all events with the css class "cal-cat-all", which our events do not have yet.
I can’t judge about that because I don’t know the details of the implementation. From the design standpoint I can only say the tick boxes should be ticked by default until the user unticks one or more.

thet:
@cornae plz check this PR: https://github.com/quaive/ploneintranet.prototype/pull/1238

PR #815, opened 2021-03-09
date picker: formatted date

thet:
pat date picker: Change styled behavior so that it can be formatted and localized while still submitting iso values.

Example:
- Example: https://github.com/Patternslib/Patterns/blob/scr-905/src/pat/date-picker/index.html#L109
- Doc in: https://github.com/Patternslib/Patterns/blob/scr-905/src/pat/date-picker/documentation.md
- Formatting options: https://momentjs.com/docs/#/displaying/format/

PRs:
- https://github.com/Patternslib/Patterns/pull/815
- https://github.com/quaive/ploneintranet.prototype/pull/1241
- https://github.com/syslabcom/oira.prototype/pull/246

Ref:
- https://github.com/quaive/ploneintranet.prototype/issues/1205
- https://jira.syslab.com/browse/SCR-905

Looks like so:
Screenshot from 2021-03-09 20-39-50

@pysailor the standard behavior - when no ``output-format`` is given, is to show ISO dates.
Note, there would be an option to automatically format the date in a localized version via this API (no IE support, though):
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
Something like:
```console.log(new Intl.DateTimeFormat('de', { dateStyle: 'long' }).format(date));```
We could actually make this variant a fallback - not sure if it's a good idea.

### DIFF
--- a/src/core/registry.js
+++ b/src/core/registry.js
@@ -149,10 +149,12 @@ const registry = {
             // - with class ``.cant-touch-this``
             // - wrapped in ``.cant-touch-this`` elements
             // - wrapped in ``<pre>`` elements
+            // - wrapped in ``<template>`` elements
             return (
                 !el.matches(".cant-touch-this") &&
                 !el?.parentNode?.closest?.(".cant-touch-this") &&
-                !el?.parentNode?.closest?.("pre")
+                !el?.parentNode?.closest?.("pre") &&
+                !el?.parentNode?.closest?.("template") // NOTE: not strictly necessary. Template is a DocumentFragment and not reachable except for IE.
             );
         });
 
--- a/src/core/registry.test.js
+++ b/src/core/registry.test.js
@@ -80,6 +80,11 @@ describe("pat-registry: The registry for patterns", function () {
                     <div class="e4 pat-example"></div>
                 </div>
             </pre>
+            <template class="e5">
+                <div>
+                    <div class="pat-example"></div>
+                </div>
+            </template>
         `;
         registry.scan(tree);
 
@@ -87,5 +92,11 @@ describe("pat-registry: The registry for patterns", function () {
         expect(tree.querySelector(".e2").textContent).toBe("");
         expect(tree.querySelector(".e3").textContent).toBe("");
         expect(tree.querySelector(".e4").textContent).toBe("");
+        expect(
+            tree
+                .querySelector(".e5")
+                .content.firstElementChild.querySelector(".pat-example")
+                .textContent
+        ).toBe("");
     });
 });
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -556,6 +556,11 @@ const jqToNode = (el) => {
     return el;
 };
 
+const ensureArray = (it) => {
+    // Ensure to return always an array
+    return Array.isArray(it) || it.jquery ? it : [it];
+};
+
 var utils = {
     // pattern pimping - own module?
     jqueryPlugin: jqueryPlugin,
@@ -581,6 +586,7 @@ var utils = {
     debounce: debounce,
     isIE: isIE,
     jqToNode: jqToNode,
+    ensureArray: ensureArray,
 };
 
 export default utils;
--- a/src/core/utils.test.js
+++ b/src/core/utils.test.js
@@ -614,4 +614,27 @@ describe("core.utils tests", () => {
             done();
         });
     });
+
+    describe("ensureArray tests", () => {
+        it("returns the passed value as is if it is an array.", (done) => {
+            const val = [1, 2, 3];
+            expect(utils.ensureArray(val)).toBe(val);
+
+            done();
+        });
+
+        it("returns jQuery object as is, as it is array-like.", (done) => {
+            const val = $();
+            expect(utils.ensureArray(val)).toBe(val);
+
+            done();
+        });
+
+        it("returns an array, if the passed value is not an array.", (done) => {
+            const val = "1, 2, 3";
+            expect(utils.ensureArray(val)[0]).toBe(val);
+
+            done();
+        });
+    });
 });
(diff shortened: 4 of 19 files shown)


pysailor:
@thet Working nicely! We will need to check if setting a date via robot tests still can be done reasonably easily (by directly setting a passed date in ISO-format to the input field, maybe?)

One part that's missing: in case a date field is not required, a user should be able to remove the date again.
I have no idea if this helps, but for pat-autosuggest, we show a little X that allows to delete a value.
@cornae Maybe we should do a quick call, the 3 of us, to spec this requirement?

Example of "remove existing value" in pat-autosuggest
Screenshot 2021-03-22 at 17 33 25

cornae:
Yes lets. Probably the best moment is after I’ve looked the OiRA date pickers myself. I have only looked at Quaive so far.

pysailor:
Hi @thet,
I updated the example page with a new scenario to demonstrate a bug: A date field that gets added via pat-clone. In this case, we get two `<time>` output fields.

thet:
@pysailor I have updated this PR with the following changes:

Clear Button:

- pre-requisite for the fix: "Core Base constructor: Await for initalization." - https://github.com/Patternslib/Patterns/pull/815/commits/3cb58141a52617e7924e3d687a8dc1966f87b05f
- The actual feature: "pat date picker: Add clear button for styled behavior to clear non-required input fields." - https://github.com/Patternslib/Patterns/pull/815/commits/1aa4a8e784005fde896235ae29e6a2db966beee3

And fixed the pat-clone issue here:

- pre-requisite for the pat-clone fix: "core dom: Add get_parents to return all parent elements from a given DOM node." - https://github.com/Patternslib/Patterns/pull/815/commits/3e4581ae26e3404a327de6f314cf8d954a36f398
- pre-requisite for tests to work: "pat dom is_visible: Mock in tests to check only for hidden to avoid unavailable offsetWidth/offsetHeight in Jest." - https://github.com/Patternslib/Patterns/pull/815/commits/94b966ea23c56b8b06d1f99fd617f5fb3d1e2508
- The actual fix: "pat date picker: Do not initialize when input field is hidden or with in a hidden tree. Avoids initializing the date picker twice when it's cloned with pat-clone." - https://github.com/Patternslib/Patterns/pull/815/commits/dc3a3ae26929946aeaeb70789a476c75b78e600a

thet:
Heads up for @pysailor and @cornae:

The pat-clone issue was because the hidden pat-clone template was already initialized by pat-date-picker. therefore it got the <time> element attached and the calender initialized.
When cloning the cloned input field was initialized again, resulting in two <time> elements.

My fix here is to not initialize pat-date-picker on input fields which are hidden - in our case the input field is hidden withing a <fieldset hidden>.

Do you think that logic is OK or can you think of any use case which would break with the assumption of hidden pat-date-picker not being initialized?

I guess it's OK and we could even think of not initializing any hidden patterns at all.

thet:
@pysailor - i noticed a ``<input placeholder..`` in your pat-clone example. placehoder are not yet supported with the new "styled" changed for pat-date-picker.
I think I'll add some support for that, right? Do we need cornelis for that or should i just copy the placeholder text into the time field?

cornae:
That’s a good one. We have to do something with it indeed. I will most likely come up with a span.placeholder inside the time attribute that I can style, but today I have no time to look at it.

thet:
@pysailor @cornae i think i know why the date picker isn't currently initialized in some cases.
These are yet just thoughts without any proof, but let's discuss this:

I'm referring to this comment above: https://github.com/Patternslib/Patterns/pull/815#issuecomment-810157688

I was implementing a logic where the date picker wasn't initialized when it's hidden, so that no two ``<time>`` elements were created - once when it's first initialized and second when it's cloned.

Now I think the date-picker might still be hidden when it's cloned. It might be a CSS animation or the collapsible in case of OIRA.

We need a different approach - preventing any patterns in pat-clone templates from being initialized.

For that I'd extend the Patterns registry to not initialize pat-clone templates when it's scanning the DOM tree.

pat-clone templates can be anywhere in the DOM tree, not only in a sub tree of the pat-clone element itself, right? That makes it a bit harder because I also need to parse the pat-clone options to get the ``template`` selector.

Is there also the case that pat-clone templates should be initialized, because they are visible? If so, that makes it even harder.

thet:
@pysailor @cornae ok, I think I have fixed the problem of double-initialization.

Now, when we add the class ``.cant-touch-this`` to the pat-clone template, then patterns within are NOT initialized.
This feature was added here: https://github.com/Patternslib/Patterns/pull/815/commits/57d86f819c3b6691456719ca1d924e62177341ac
And documented for pat-clone here: https://github.com/Patternslib/Patterns/pull/815/commits/ec368ddf70df75c5e8c61720cf8902ea0a31b97a


I briefly followed the idea of using ``<template>`` elements in https://github.com/Patternslib/Patterns/pull/824 but that needs a little more work in pat-clone to support that.

cornae:
The class sounds good to me. Couldn't we just as well just skip elements that have a hidden attribute on it or that are display none-ed?

thet:
@cornae you're right - did that.
that way we probably do not need any changes in our pat-clone templates because they already have the ``hidden`` attribute.

pysailor:
Looking good. Some findings:

* If there is a validation error (Start must be before End) and I then delete either start or end, I expect the validation to be gone (since there is nothing to compare any more). Does not happen currently
* Validation errors in `<em>` should be below the input and time fields (as it was before). Now they appear above the `<time>` field.
* When there is a data-pat-date-picker option "after" to let a date field automatically increment, we get a validation error like this:
Screenshot 2021-04-07 at 15 19 27

PR #854, opened 2021-06-10
pat-tabs: Extra classes to indicate tabs state

### LINKED ISSUE #841: Extra classes for pat-tabs
- [ ] A class `tabs-wrapped` that appears on the pat-tabs element when a span was wrapped around items that didn't fit.
- [ ] A class `tabs-ready` that appears on the pat-tabs element as soon as pat-tabs is done with its transformation.

Both classes appear on the element that has .pat-tabs on it.

.tabs-wrapped is the essential one. If we have .tabs-wrapped, then .tabs-ready is the nice to have.

reinhardt:
Note: The tests currently don't cover resizing the browser after first init; e.g. when width increases the `tabs-wrapped` class may need to disappear. I've verified this manually only.

Closes #841

### DIFF
--- a/src/pat/tabs/tabs.js
+++ b/src/pat/tabs/tabs.js
@@ -39,6 +39,12 @@ export default Base.extend({
     },
 
     adjust_tabs() {
+        this.el.classList.remove("tabs-ready", "tabs-wrapped");
+        this._adjust_tabs();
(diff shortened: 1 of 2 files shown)


thet (approved):
LGTM

PR #855, opened 2021-06-10
pat-scroll: Allow a delay before scrolling

### LINKED ISSUE #842: Delay property for pat-scroll
The same delay property that we have for other patterns also for pat scroll. Example use case: scrolling to a child of a collapsible while opening the collapsible at 'the same time'. 

The delay would allow for the collapsible to be opened before pat scroll does it's position calculation.

Important detail is that the calculation of how much pat-scroll will scroll happens after the delay.

We have property early already for pat-inject and pat-forward.

reinhardt:
~~Hold merge; I'm not yet clear on what the issue is (see #842) so I don't know whether this solves it.~~

Closes #842

### DIFF
--- a/src/pat/scroll/index.html
+++ b/src/pat/scroll/index.html
@@ -116,12 +116,21 @@
                 >
             </p>
         </article>
-        <div class="pat-well pat-collapsible closed pat-scroll"
-             data-pat-collapsible="transition: fade; effect: slow;"
-             data-pat-scroll="selector: #s2; trigger: click"
+        <p>
+          The “Open” button opens the collapsible below and scrolls to the second section.
+        </p>
+        <fieldset class="pat-button-bar">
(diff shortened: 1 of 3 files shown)


cornae:
@reinhardt The grooming for this ticket was done with @thet. If you are picking up this ticket then I propose to do the grooming again with you. Having extra layers of separation within the teams increases the chance on misunderstandings and delays.

reinhardt:
I'm not sure what exactly the “grooming” involves but I'm available to clarify in a call if you like!

cornae:
The grooming is discussing the ticket with it's intention and all the nuances with the team. It was done with only @thet in the case assuming that @thet would do this ticket. I would happily do the grooming again, but it's probably not needed anymore :).

thet (approved):
LGTM

Issue #847, opened 2021-05-25
Pat-bumper class bumped


cornae:
Pat-bumper places a class `bumped` when the element bumps. It only places this class when the element bumps to the top of the first scrolling parent. The proposal of this ticket is to also set the class when it bumps to the bottom of the first scrolling parent.

thet:
Closed via: https://github.com/Patternslib/Patterns/pull/857

PR #857, opened 2021-06-11
pat-bumper: Extra classes depending on bumped side

reinhardt:
This is a little lazy - the main class is configurable but the new classes are hard coded.

Refs #847


thet (approved):
LGTM

PR #861, opened 2021-06-14
Rework pat-bumper

thet:
breaking(pat bumper): Rework pat-bumper to onlys set bumping classes. positions are not set, do that with position:sticky.

Based on
- https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
- https://davidwalsh.name/detect-sticky


thet:
@reinhardt this does not work yet, although there should not much be left to get it working. Let's discuss this tomorrow.

PR #877, opened 2021-07-09
feat(pat push): Added the option to append the fetched content at the…

### LINKED ISSUE #876: push: enhance the pattern allowing to append the the fetched content to the .pat-push element
Use case:
- we have a chat window open with a lot of messages
- a new message arrive
- I want to see it pushed into my page

The current implementation of pat-push does not allow that because it will replace the content of the element all together.

ale-rt:
… end of the pat-push element

Fixes #876

### DIFF
--- a/src/pat/push/documentation.md
+++ b/src/pat/push/documentation.md
@@ -7,3 +7,5 @@ When an element is equipped with class="pat-push" and a push ID it will listen t
    data-pat-push="push-id: message_counter; url: /message-counter">
 Messages <sup class="counter digit-1">1</sup>
 </a>
+
+If you add the `mode: append` option, the fetched content will be appended to the element.
--- a/src/pat/push/index-push-messages.html
+++ b/src/pat/push/index-push-messages.html
@@ -1,0 +1,1 @@
+<li>A <strong>new</strong> message</li>
--- a/src/pat/push/index.html
+++ b/src/pat/push/index.html
@@ -42,5 +42,18 @@
         >
             messages <sup class="counter digit-1">1</sup>
         </div>
+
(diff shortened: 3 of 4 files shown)


cornae:
This change is not part of the pat push pattern design and has also not been discussed in the team.

ale-rt:
Well, I do not see any reason for this to require a brainstorming session it is a no brainer if you check the use case described in #876.
This is very demovating.

cornae:
I read that ticket. I see many reasons @ale-rt that you don't see while there is no reason why we need this 'no brainer' now and I can explain why but not running behind pull requests, but rather in a proper team session instead. And even if there were no reasons, we have to follow the process to keep everybody on board. Also, I want to avoid the possibility of another disaster striking upon the user like what happened with the events sidebar hack.

ale-rt:
So many interesting stories to learn about

thet:
As part of getting a necessary feature ready to demo we need this PR to be merged.
Note: this is a technical necessity. We need to gain some knowledge about using pat-push in real environments until we can come up with better approaches.
This whole pattern is very technical and doesn't work without a proper push_kit / RabbitMQ setup anyways and is yet not promoted on the patternslib.com website. We always can fix any issues with it in upcoming versions (following semantic versioning which we do, major versions can always break backwards compatibility).

We discussed another approach of using pat-inject. This does not work in this case, as pat-push isn't using pat-inject at all.

However, this PR seems fine to me.

thet (commented):
LGTM

thet:
Maybe there was also a confusion about pat-inject push-marker. This is a different story, this PR is about pat-push.

PR #880, opened 2021-07-09
maint: Fix the pre-commit command

### LINKED ISSUE #879: Pre commit does not work as expected
I see hhusky is installed with `make stamp-yarn` https://github.com/Patternslib/Patterns/blob/6703655ee13ae8cf42dee7e084e5afc511bf3d54/Makefile#L24

But:
```
ale@code:~/Code/javascript/Patterns$ make stamp-yarn 
make: 'stamp-yarn' is up to date.
ale@code:~/Code/javascript/Patterns$ git commit -a --amend -m "test"
.husky/commit-msg: 4: yarn: not found
husky - commit-msg hook exited with code 127 (error)
```

ale-rt:
I am not sure about this patch, but it seems to work for me.

Fixes #879

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-./node_modules/.bin/yarn commitlint --edit $1
+npx yarn commitlint --edit $1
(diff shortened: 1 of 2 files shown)


thet (approved):
LGTM and should work.
However, I did change the command to use ``npx`` which is a node standard tool to auto-discover or even auto-install a node program in case it's not available.

PR #893, opened 2021-09-08
Add remove events

thet:
- feat(core dom): add_event_listener and remove_event_listener methods to register event handlers for DOM nodes which can be unregistered by an id. Event handlers with the same id on the same node won't be registered twice.

Plus:
- Upgrade dev dependencies up to minor releases and other releases up to patch level. /pt2.
- feat(core base): Store the pattern instance on the element also when instantiating manually without scanning the DOM. Also store the pattern instance directly on the DOM node without jQuery.data.
- fix(pat modal): Wait a tick before destroying the modal so that registered event handlers (like form submit) can kick in before the modal disappears.


pilz:
I have no clue what or why you are doing this, besides what you explained on a high level today, but I fully trust you :)

PR #894, opened 2021-09-08
pat-inject error handling: Fallback to rendered template from error response.

thet:
feat(pat inject): Error handling: In a case of injection error, search for a configurable CSS selector in the error response. If it is found, inject the rendered error page into the document. If not, fall back to pre-configured error pages.

## Related PRs
https://github.com/quaive/ploneintranet/pull/4350
https://github.com/quaive/quaive.resources.ploneintranet/pull/61
https://github.com/quaive/ploneintranet.prototype/pull/1406
https://github.com/Patternslib/Patterns/pull/894

### DIFF
--- a/src/pat/inject/inject.js
+++ b/src/pat/inject/inject.js
@@ -626,7 +626,6 @@ const inject = {
 
     async _onInjectError($el, cfgs, event) {
         let explanation = "";
-        let fallback;
         const status = event.jqxhr.status;
         const timestamp = new Date();
         if (status % 100 == 4) {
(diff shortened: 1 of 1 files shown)


src/pat/inject/inject.js
```diff
@@ -640,21 +639,41 @@ const inject = {
...
         }
 
+        let fallback;
         const url_params = new URLSearchParams(window.location.search);
+        if (url_params.get("pat-inject-errorhandler.off") === null) {
+            // Prepare a error page to be injected into the document.
+
+            // Check if a valid error page can be found in the response.
+            const valid_error_page_selector = document
+                .querySelector(`meta[name=pat-inject-valid-error-page]`)
```
thet, line 649: This introduces a new error handling meta tag configuration parameter: ``meta[name=pat-inject-valid-error-page]``

cornae:
Please hold merge.

src/pat/inject/inject.js
reinhardt: Why do we need this parameter? I imagine it is to exclude a case where some HTML is returned in case of an error, but we don't want to use it. What would such a case look like? Is it not reasonable to assume that any HTML that is returned with an error status is indeed an error page?

pilz: If an error happens, it can be
- just a one liner text string ("Error") because a programmer just returned a string :)
- an html page generated by some backend service like solr, haproxy, nginx, varnish (The famous Guru meditation error)
- an error page from plone that is unstyled

In all these cases we do not have a markup structure that has a valid source for injection and hence it would fail to inject the error. We want to be able to detect that and fall back to a standard error page - but inject, if the error is usable.

pilz:
@cornae can you quickly tell why you put the hold merge? Do you need info why this is necessary? Do you have a technical concern?

cornae:
Because at that moment the topic was undiscussed while it talks about it talks about solutions how injections are handled.

pilz:
Ok, cool. Now we have come so far that inject is so smart to inject the real styled plone error page in case one is presented. pat-inject just needs to know if the error markup is usable (styled) or some technical mumblejumble, and thet is using the css selector to detect that. This is an improvement over showing the generic error from the assets page and should not harm, because it only activates if plone presents a proper error page. 
With that explanation, do you think we can merge this?

pilz:
This ticket is a candidate for Quaive 3.2 because it would improve the current error page situation considerably, showing proper errors when available to users.

cornae:
I have many questions, each answer to these questions might generate more questions. Shall we save time and book a meeting for this? Perhaps it can be combined with the meeting on error pages that you were still going to organise.

pilz:
@cornae Yes, wonderful. Can you do today? Even now?

pilz:
Had the alignment call. We need to revisit this ticket in combination with the outcome of the ticket https://github.com/quaive/ploneintranet.prototype/issues/1423. Once 1423 is designed, we take this one further.

src/pat/inject/inject.js
reinhardt: > * just a one liner text string ("Error") because a programmer just returned a string :)

No, not really. If the programmer lets a view return a string, the merit can be debated, but then that's the success case. If an error happens, you get a traceback wrapped in the Plone error page. To prevent that, the programmer would have to catch any and all errors and explicitly suppress the exception traceback, returning a string instead. I doubt that it makes sense to consider that a possible case.

> * an html page generated by some backend service like solr, haproxy, nginx, varnish (The famous Guru meditation error)

Solr responses are never directly handed to the client. The rest sounds plausible.

> * an error page from plone that is unstyled
> 
> In all these cases we do not have a markup structure that has a valid source for injection and hence it would fail to inject the error.

I don't know of any markup that could not in principle be injected. Could it be that your concern is that we get an unstyled error, i.e. the injection would not technically fail but the result would be visually “wrong” in some way?

PR #914, opened 2021-10-28
pat-calendar: Default to 24 hours time format

reinhardt:
Fixes quaive/ploneintranet#3658


reinhardt:
`time-format` is indeed currently completely ignored by `pat-calendar`. Restoring it is not possible because the FullCalendar interface has changed. #915 tries to implement a replacement.

PR #915, opened 2021-10-28
pat-calendar: Fix `time-format` parameter

reinhardt:
`time-format` was ignored by pat-calendar prior to this PR. It probably used to map to `timeFormat` which does not exist any more in FullCalendar version 5. The new eventTimeFormat has a different spec, see https://fullcalendar.io/docs/eventTimeFormat


reinhardt:
@thet, this makes it possible to set the time format to 24h regardless of the language/locale. I'll tweak and rebase it and then pass it to you for review.

reinhardt:
Here we go!

reinhardt:
Proposing to merge #933 instead

PR #922, opened 2021-11-10
[WIP] pat-modal: rework event registration to not multi-run handlers

thet:
Note: Since this changes event handler registration from jQuery to plain JavaScript this PR could possibly break things.
This should go into the next major release where more testing is needed anyways.


thet:
Replaced by: https://github.com/Patternslib/Patterns/pull/920

PR #933, opened 2021-12-15
New format-time parameter

reinhardt:
Possible values: `12h`, `24h`; default: `24h`. See quaive/ploneintranet#3658

Closes #915


thet (approved):
Thanks, looks very good to me!
I'm currently to add little addition so that with invalid configuration (e.g. old-style format) it will also use the 24h option. Then I'll merge. OK?

PR #943, opened 2021-12-22
pat-inject modernization

thet:
A side outcome of the pat-inject autoload-visible fixes.

Eventually this will be revisited later to rework pat-inject for a more readable, simpler and clearer structure.

The branch name is misleading.


thet:
This is replaced by: https://github.com/Patternslib/Patterns/pull/1222
Except for the Base prototype rework, which is obsolete on it's own since I'd use the class based pattern approach now.
However, reworking into a class-based pattern is intended as it will give us encapsulation. Now we cannot bind any variables on `this` but need to bind everything on the node element itself. This will make the code simpler, more readable and more error-prone.

PR #984, opened 2022-03-23
testing ajax-caching errors

thet:
…che which can be set to caching or no-caching.


cornae:
A nice improvement of pat-inject that might also be useful in other places in the future.

thet:
btw - i'm testing in this branch a test failure which occurs on https://github.com/Patternslib/Patterns/pull/981 and which i do not yet understand...

PR #1010, opened 2022-05-23
pat-modal: allow modal in modal

thet:
Changes:

- Add new option ``target`` for pat-modal which is passed to pat-inject when the modal injects content. This allows you to open modals in modals by selecting the already existing parent modal as injection target: ``#pat-modal::element``.

- Adding demo to show how to add a modal-in-modal.

- The class close-panel should not be added on the child-modal link.

@cornae this is a draft solution for further discussion.

I don't think modal-in-modals were supported until now. My fist attempt fixing .close-panel to allow  a child modal to be injected and instantiated while closing the parent modal failed, because the child modal is opened within the parent one. I cannot close the parent modal without loosing the child modal.
We could fix this if we would always open modal on the body root. But this is not the case yet.

However this approach outlined in this PR seems to work.

We might change the pattern option name ``target`` to your liking and I probably have to optimize the code a bit more (unregistering event handlers, etc.).


thet:
TODO:
- add ``source``.
- actually, rethink this whole process - currently we need an extra template, which replicates an already pre-built modal which then replaces the parent modal via pat-inject. We actually want the child modal to have the same markup structure than a parent modal.

I'll rethink this whole thing

thet:
This is an unobtrusive change and allows to open modals in modals immediately.
If there is a need for further optimizing this feature, we can do so - later.

PR #1075, opened 2022-09-28
fix(pat navigation): Don't check query string and hash fragments.

thet:
When setting navigation markers based on URL comparison do not set the query string and hash fragments. When using hash navigation within one page, use pat-scroll to mark items as current.

This fixes a problem described in: https://github.com/euphorie/Euphorie/pull/447#issuecomment-1259660758

It has the drawback that URL based detection of the current item when using navigation based on query strings or hash fragments (e.g. a table of contents) would not be possible anymore. But that might be corner cases.

### DIFF
--- a/src/pat/navigation/navigation.js
+++ b/src/pat/navigation/navigation.js
@@ -207,7 +207,12 @@ export default Base.extend({
      * @returns {String} - The prepared url.
      */
     prepare_url(url) {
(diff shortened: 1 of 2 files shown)


reinhardt:
Did you consider Ale's suggestion of making this configurable?

thet:
Will do that

PR #1088, opened 2022-11-24
feat(pat-autosuggest): Add a configurable separator for multiple values.

thet:


src/pat/auto-suggest/documentation.md
```diff
@@ -38,16 +38,17 @@ Pre-fill the input element with words in JSON format and don't allow the user to
...
+| prefill              | List    |               | A comma separated list of values with which the form element must be filled in with. The `value-separator` option does not have an effect here.                                                                                                                                      |
+| prefill-json         | JSON    |               | A JSON object containing prefill values. We support two types of JSON data for prefill data:`{"john-snow": "John Snow", "tywin-lannister": "Tywin Lannister"}` or `[{"id": "john-snow", "text": "John Snow"}, {"id": "tywin-lannister", "text":"Tywin Lannister"}]`                  |
+| value-separator      | String  | ","           | A separator character to separate multiple values from each other. This is used to initialize and write back the selection to the input field. The prefilled option is always separated by a comma. If you want to use a semicolon as separator do it like so: `value-separator: ;;` |
```
reinhardt, line 51:
```suggestion
| value-separator      | String  | ","           | A separator character to separate multiple values from each other. This is used to initialize and write back the selection to the input field. The `prefill` option is always separated by a comma. If you want to use a semicolon as separator do it like so: `value-separator: ;;` |
```

src/pat/auto-suggest/auto-suggest.js
```diff
@@ -50,6 +51,7 @@ export default Base.extend({
         this.options = parser.parse(this.el, this.options);
 
         let config = {
+            separator: this.options.valueSeparator,
```
ale-rt, line 54: I do not get why we have a `separator` and a `tokenSeparator`. I hope it's fine like that :)

thet: initially I thought we need to modify the `tokenSeparator`. But not, it's for the tagging feature - when you add the tokenSeparator a new tag/token will be created. See:
https://select2.github.io/select2/#doc-tokenSeparators

ale-rt (approved):
LGTM

PR #1106, opened 2022-12-07
New Modernizr loader

thet:
Replace the `webpack-modernizr-loader` which targets Webpack 4 with https://github.com/webpack-contrib/val-loader#modernizr

This is on hold now because the `js`/`no-js` classes are not set and we need especially those.


thet:
Got obsolete by: https://github.com/Patternslib/Patterns/pull/1115

PR #1126, opened 2023-01-23
Feature: input checkbox to toggle checked boxes true/false.

petschki:
I need a checkbox (for example at the header column of a table) to select/deselect all sibling checkboxes.

This extends the `pat-checklist` pattern to look for a `<input class="toggle-all" />` which selects/deselets all its siblings according to its checked state.

### DIFF
--- a/src/pat/checklist/checklist.js
+++ b/src/pat/checklist/checklist.js
@@ -26,7 +26,9 @@ export default Base.extend({
     },
 
     _init() {
-        this.all_checkboxes = this.el.querySelectorAll(`input[type=checkbox]:not(${this.options.toggle}`);
+        this.all_checkboxes = this.el.querySelectorAll(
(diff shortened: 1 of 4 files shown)


petschki:
https://user-images.githubusercontent.com/511761/214022309-dfd1897a-bb99-4042-a50e-a1569237f7c2.mov

thet:
@petschki Thanks for your PR!

The PR‌ #965 was to fix pat-toggle - which only toggles a class on an element - to work together with pat-checklist and not steal pat-checklist the click event. before, toggling did work, but the checkbox wasn't checked/unchecked.

So you PR makes sense.

@cornae do you have any concerns about the new pat-checklist argument `toggle` which is a CSS selector to find toggle-button nad defaults to `.toggle-all`?

image

thet (approved):
LGTM

PR #1157, opened 2023-05-10
fix(pat calendar): Do not set a Content-Type header when no body is submitted

### LINKED ISSUE #1156: Is the Content-Type header needed in pat-calendar requests
We are currently facing an issue with Zope refusing to respond to pat-calendar ajax calls, see https://github.com/plone/plone.restapi/issues/1611.

I think the Content-Type header is not needed here:
https://github.com/Patternslib/Patterns/blob/7e391d33ed84bf8461d7c0c04f80ec157c23e549/src/pat/calendar/calendar.js#L311-L318

ale-rt:
Closes #1156

### DIFF
--- a/src/pat/calendar/calendar.js
+++ b/src/pat/calendar/calendar.js
@@ -313,7 +313,6 @@ export default Base.extend({
                     mode: "cors",
                     headers: {
                         "Accept": "application/json",
-                        "Content-Type": "application/json",
                     },
                 });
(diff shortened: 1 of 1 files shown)


ale-rt:
Note that there is another occurrence of `Content-Type` in the file, but in that case a body was submitted.

thet (approved):

PR #1160, opened 2023-05-16
pat-inject scroll fix

thet:
- fix(pat inject): Fix a JS error with pat-inject where the scrolling target could not be found.

- fix(pat inject): Allow to use a scroll container other than the injection target.
This fixes a problem when the content should scroll to an element which is within a scroll container below the injection target. Until now the scroll container could only be a parent of the injection
target. Now it can be a parent of the scroll-target.


thet (commented):
This is probably the real fix for the JS error, if I'm not mistaken.
The other stuff is some modernization code while analyzing and trying to reproduce and understand the problem.

PR #1161, opened 2023-05-17
maint(pat inject): Modernize some parts of pat inject.

thet:


thet:
Replaced by: https://github.com/Patternslib/Patterns/pull/1222

PR #1163, opened 2023-05-17
fix(pat validation): Do not disable input elements with formnovalidate.

### LINKED ISSUE #1132: Submitting buttons with `formnovalidate` doesn't work, because they are disabled.
The fix here https://github.com/Patternslib/Patterns/issues/960 doesn't work. We never get to this line, because the buttons with `formnovalidate` parameter get disabled before.

https://github.com/Patternslib/Patterns/blob/master/src/pat/validation/validation.js#L121

thet:
Do not disable input elements with the `formnovalidate` attribute set when form validation fails.
E.g. a cancel button: `<button formnovalidate>Cancel</button>`.

Fixes #1132.

### DIFF
--- a/src/pat/validation/validation.js
+++ b/src/pat/validation/validation.js
@@ -117,7 +117,7 @@ class Pattern extends BasePattern {
         const validity_state = input.validity;
 
         if (event?.submitter?.hasAttribute("formnovalidate")) {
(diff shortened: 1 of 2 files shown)


petschki (commented):
approved after merge 😄

PR #1195, opened 2024-01-03
feat(pat-autosuggest): load more

thet:
syslabcom/scrum#1638


src/pat/auto-suggest/auto-suggest.js
```diff
@@ -258,14 +260,28 @@ export default Base.extend({
...
                                 page: page,
                             };
                         },
                         results: (data, page) => {
-                            // parse the results into the format expected by Select2.
+                            // Parse the results into the format expected by Select2.
                             // data must be a list of objects with keys "id" and "text"
-                            return { results: data, page: page };
+
+                            // Check, if there are more results to come.
```
reinhardt, line 271:
```suggestion
                            // Check whether there are more results to come.
```

pilz:
I noted on 11.June that @cornae wanted to contact @thet on the specifics of this PR to determine how to name the pattern attributes. Could you do this this week? The timeouts we get are causing repeated helpdesk requests from osha and are getting quite disruptive, so fixing this will have an immediately positive impact on our peace.

thet:
@reinhardt ready for review!

src/pat/auto-suggest/documentation.md
```diff
@@ -34,16 +34,52 @@ Pre-fill the input element with words in JSON format and don't allow the user to
...
+
 ### Option reference
 
 You can customise the behaviour of a gallery through options in the `data-pat-auto-suggest` attribute.
 
 | Property             | Type    | Default Value | Description                                                                                                                                                                                                                                                                          |
 | -------------------- | ------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| ajax-batch-size      | Number  | 10            | Batch size for subsequent pages of a bigger batch. For the first page, `max-initial-size` is used.                                                                                                                                                                                   |
```
reinhardt, line 76:
```suggestion
| ajax-batch-size      | Number  | 10            | Batch size for subsequent pages of a bigger result set. For the first page, `max-initial-size` is used.                                                                                                                                                                                   |
```

src/pat/auto-suggest/documentation.md
```diff
@@ -34,16 +34,52 @@ Pre-fill the input element with words in JSON format and don't allow the user to
...
 | ajax-data-type       | String  | "json"        | In what format will AJAX fetched data be returned in?                                                                                                                                                                                                                                |
 | ajax-search-index    | String  |               | The index or key which must be used to determine the value from the returned data.                                                                                                                                                                                                   |
+| ajax-timeout         | Number  | 400           | Timeout before new ajax requests are sent. The default value is set ot `400` milliseconds and prevents querying the server too often while typing.                                                                                                                                   |
```
reinhardt, line 79:
```suggestion
| ajax-timeout         | Number  | 400           | Timeout before new ajax requests are sent. The default value is set to `400` milliseconds and prevents querying the server too often while typing.                                                                                                                                   |
```

src/pat/auto-suggest/auto-suggest.js
```diff
@@ -11,7 +11,10 @@ const log = logging.getLogger("autosuggest");
 export const parser = new Parser("autosuggest");
 parser.addArgument("ajax-data-type", "JSON");
 parser.addArgument("ajax-search-index", "");
+parser.addArgument("ajax-timeout", 400);
 parser.addArgument("ajax-url", "");
+parser.addArgument("max-initial-size", 10); // AJAX search results limit for the first page.
+parser.addArgument("ajax-batch-size", 10); // AJAX search results limit for subsequent pages.
```
reinhardt, line 17: In https://github.com/Patternslib/Patterns/pull/1195#issuecomment-2208705609 it says that the default for `ajax-batch-size` is supposed to be 0, in which case IIUC no batching is performed. Is there a reason that it is done differently here?

reinhardt (changes requested):

src/pat/auto-suggest/auto-suggest.js
thet: fixed!

thet:
@reinhardt finally implemented the specs as defined (did not carefully enough read them... :/ )

However, there is one case not covered by the specs - loading all items at once on the first page. Is this a use case we want to support?

Currently the code supports it:

- If you set `max-initial-size` to `0` all the items will be loaded on the initial page.
- On the other hand `ajax-batch-size` set to `0` will disable batching and only show the first page.

I think allowing to load all the items from the server at once without defining the `max-initial-size` to some ridiculous high number could be useful. Maybe a `-1` would be better?
@cornae any input on that? If you want, we can discuss that tomorrow or handle it sometime in the future - it's not documented, so this behavior doesn't exist officially.

reinhardt:
I don't have any strong opinions about any of it. I was just wondering why it was different than described.

That said, it might be a good idea to have a limit on the number of results. The reason we are working on this is, after all, that we got in trouble because there were too many entries.

On the other hand, a `max-initial-size` equal to 0 is a degenerate case - why would you want to do ajax but not be interested in any results? So I'm also fine with giving it a special meaning. You shouldn't set it to 0 without knowing what you're doing anyway.

thet:
According to a chat with @cornae we can keep this PR as it is. Also `max-initial-size` set to `0` and loading all elements, as long this is not documented.
We might improve it further in the future, if needed.

PR #1199, opened 2024-07-23
Update CHANGES.md

pilz:


ale-rt:
I expected the changelog to be compiled by some tool because of the semantic commit messages :/ Maybe I did something wrong

PR #1207, opened 2024-10-24
fix(pat-inject): Allow to submit invalid forms when form/novalidate.

thet:
Respect the form[novalidate] and button[formnovalidate] attributes and allow the submission of invalid forms.
This is necessary if a cancel button is a submit button and needs to be allowed to submit even if the form input is invalid.

Note: there is one skipped test due to lack of `formnovalidate` support in jsDOM. However, I have tested it manually on a testing page. The fix seems to work in Firefox.

TODO:
- [ ] Create an issue for jsDOM for missing `button[formnovalidate]` support.
- [ ] Even fix this in jsDOM.

### DIFF
--- a/src/pat/inject/inject.js
+++ b/src/pat/inject/inject.js
@@ -186,14 +186,20 @@ const inject = {
         const $el = $(e.currentTarget);
         let cfgs = $el.data("pat-inject");
         if ($el[0].tagName === "FORM" && e.type === "submit") {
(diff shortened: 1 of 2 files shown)


petschki (approved):
LGTM

mauritsvanrees (approved):
I'm not sure how to test this on my machine in the project where I saw this problem.  But the change looks logical.
Thanks!

PR #1208, opened 2024-12-12
feat(pat-date-picker): Cache the ajax call to retrieve i18n picker translations.

thet:
Ref: scrum-2726

This implementation uses good ol' `store` module from `@patternslib/patternslib/src/core/store` - it's still handy.
However, we can spare us the `supported` query because the web storage is just supported in practice by all modern browsers since Firefox 2 etc.

Testing FTW, I found a case where a unsuccessful call to an i18n URL led to an error.

NOTE: in a real environment where multiple instances are
initialized at once on the same page, before the ajax call has
been completed, each instance will do an AJAX call. After that,
when navigating to other pages with other date picker instance
the cached value should be used and no more AJAX calls should be
made.

It would be hard to work around this limitation and I think it's a tradeoff we can live with, right?

### DIFF
--- a/src/pat/date-picker/date-picker.js
+++ b/src/pat/date-picker/date-picker.js
@@ -5,6 +5,7 @@ import logging from "../../core/logging";
 import Parser from "../../core/parser";
 import dom from "../../core/dom";
 import events from "../../core/events";
+import store from "../../core/store";
 import utils from "../../core/utils";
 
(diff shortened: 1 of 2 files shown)


thet:
Done in: https://github.com/Patternslib/Patterns/pull/1208

reinhardt (approved):

PR #1209, opened 2024-12-31
feat(pat-depends): Support for required fields.

thet:
If a dependend field is required but not shown, remove the required attribute.
Restore it once the dependend field is shown.

Ref: scrum-2615

Current state: automated tests, plausability check and battle testing missing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     },
     "maintainers": [
         {
-            "name": "Syslab.com GmbH",
+            "name": "syslab.com GmbH",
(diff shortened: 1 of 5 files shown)


thet:
Obsolete since https://github.com/Patternslib/Patterns/pull/1210

A disabled input does not participate in form validation nor is it submitted. The `required` attribute can stay as it is.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

PR #1231, opened 2025-01-28
fix(core feature-detection): Fix loading of modernizr script.

thet:
1) fix(core feature-detection): Fix loading of modernizr script.

In some situations loading of the modernizr.min.js script failed because the base url could be incorrect. That is fixed with a non-IE compatible method of getting the current's script URL, which is safe to use. The problem surfaced in Plone while loading the Patternslib library and in between the protect.js script was loaded. The base url was calculated for the protect.js script and the modernizr script could not be loaded.

2)‌ maint(core log): Console.debug is not deprecated. Just use it.

3) fix(core registry): Do nothing with Patterns without a trigger.
    Patterns without a trigger broke the registry scan method. Now they don't.

### DIFF
--- a/src/core/feature-detection.js
+++ b/src/core/feature-detection.js
@@ -18,15 +18,14 @@
     }
 
     // Get the current script tag's URL.
-    // See: https://stackoverflow.com/a/984656/1337474
-    const scripts = document.getElementsByTagName("script");
-    const script = scripts[scripts.length - 1];
-    let script_url = script.src;
+    const script_url = document.currentScript.src;
     // Get the base URL of the current script tag's URL.
-    script_url = script_url.substring(0, script_url.lastIndexOf("/")) + "/";
+    let base_url = script_url.substring(0, script_url.lastIndexOf("/")) + "/";
+    // The modernizr script is located outside the chunks directory.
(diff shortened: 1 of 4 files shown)


thet:
@cornae sounds good, I'm :+1: for phasing that out.
Maybe the 9.10 series would be a good fit.
If we agree on that I could make another alpha after that one and remove Modernizr.

I will still proceed an make an alpha version with this fixes.

thet:
@cornae there is also this fix:

"fix(core registry): Do nothing with Patterns without a trigger."

I found it while integrating the whole Patternslib bundle into Plone.
The source of the error was the `focus` pattern. This one does not have any trigger set, but initializes by it's own. So I think that the `focus` patterns still works. Please test it, if you have a usecase.

However, I think we can easily deprecate / remove the focus pattern.
If it's purpose is to set a class if one of the child elements has a focused state, we can nowadays do it like so:

```css
:has(:focus) {}
```
https://developer.mozilla.org/en-US/docs/Web/CSS/:has

or even like so (as I just found out):

```css
:focus-within {}\
```
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within

PR #1248, opened 2025-05-09
Some performance optimizations....

thet:


thet:
Included in: https://github.com/Patternslib/Patterns/pull/1249

PR #1264, opened 2025-09-09
Add some fallbacks for scrolling mechanisms

thet:
Not sure, if this is even needed. Was fixing a problem, which is not a problem anymore.


thet:
:tada:

PR #1281, opened 2026-04-15
Use long commit hash to simplify pnpm installations with patternslib dependency

petschki:
`pnpm` doesn't accept short commit hashes so we had to define an override here: https://github.com/plone/mockup/blob/master/package.json#L127


thet (approved):
LGTM

PR #1288, opened 2026-06-03
Signals polyfill + pat-bind two-way databinding pattern

thet:
This adds the new `pat-bind` pattern for two-way data binding using the not-yet-ready-specified TC39 signals specification in a custom polyfill.

pat-bind allows for declarative two-way data binding:

```html
<select class="pat-bind" data-pat-bind="key: color">
  <option value="red">Red</option>
  <option value="green">Green</option>
</select>

<span class="pat-bind" data-pat-bind="key: color; value: ::text"></span>
```

__The full documentation is at: https://github.com/Patternslib/Patterns/blob/thet/signals/src/pat/bind/documentation.md__

NOTE: event TC39 signals is not ready yet, it is used in preact, lit. The specification is meant as a foundation for any reactive framework out there.

https://github.com/tc39/proposal-signals
https://github.com/proposal-signals/signal-polyfill
https://github.com/maverick-js/signals
https://github.com/preactjs/signals
https://lit.dev/docs/data/signals/

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,6 +24,12 @@
                     >Inject history<sup>unknown</sup></a
                 >
                 <a
+                    class="navigation bind"
+                    href="/src/pat/bind/index.html"
+                    title="Bind"
+                    >Bind<sup>β</sup></a
+                >
+                <a
                     class="navigation switch"
                     href="/src/pat/switch/index.html"
                     title="Switch"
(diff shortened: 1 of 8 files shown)


MrTango:
That sound interesting. Signals are used by many JS frameworks already, Svelte 5 introduced them under the name runes. 
They do a bit moire though, besides simple binding ($state) we have there also ($derived), to declare dependencies between states.

PR #1291, opened 2026-09-10
PoC: wait for Module Federation remotes before the initial scan and implement `replace` option

petschki:
Module Federation remote bundles (add-ons) register their patterns and components asynchronously after document ready, while the registry scanned the DOM right at document ready. Whether an add-on's pattern override or component registration was in place for the initial scan was a race between the remote's chunk loading and the host's own initialization — usually won by the remote, but not guaranteed.

``registry.init()`` now awaits ``window.__patternslib_mf_initialized``, the promise provided by the Module Federation helper of @patternslib/dev, before the initial scan. Without a Module Federation host on the page nothing changes. A timeout (default 5000ms, ``window.__patternslib_mf_init_timeout``) makes sure a broken or slow remote cannot block the page; the scan then runs anyway with a warning. Repeated ``init()`` calls while waiting are ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,9 @@ global settings or access otherwise hidden objects.
 | window.\_\_patternslib_patterns_blacklist             | A list of patterns that should not be loaded.                        | []        |
 | window.\_\_patternslib_registry                       | Global access to the Patternslib registry object.                    | -         |
 | window.\_\_patternslib_registry_initialized           | True, if the registry has been initialized.                          | false     |
+| window.\_\_patternslib_registry_initializing          | True, while the registry waits for Module Federation remotes before the initial scan. | undefined |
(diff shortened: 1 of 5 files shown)


petschki:
see analysis and usecase here https://github.com/plone/mockup/pull/1637

thet (approved):
Also looks very good to me and solves a registration problem very elegantly.

Up until now, a pattern from an add-on remote bundle called `registry.register` to register the pattern and re-run a dom scan when the scan was already done. Now - with everything in place (also for the enty point's `export default import("./bundle");` trick - undocumented yet, but there is not much documentation anyways), everything is done only once and there is a better way to override existing patterns.

Sounds like a big win for me and a more stable overriding situation.

Approved!

